PR #86, opened 2015-01-07
Added blank identifier explanation to maps

### LINKED ISSUE #85: The blank identifier is explained in the Range section yet used in a previous section
The blank identifier is explained [in the range section.](https://gobyexample.com/range)
Yet it is used in the previous [Maps section](https://gobyexample.com/maps) without explanation. The blank identifier should be explained in the Maps section. 

I'm happy to put together a PR for this.

mathonsunday:
This will resolve #85.


mmcgrana:
Hi @mathonsunday. Thanks for pointing out #85 and suggesting a patch.

I'd like to tweak a few things before merging this:

* We'll need the source as well as the generated output for the Maps page (we only have the latter in this PR). If it's unclear how to build the site from source, let me know and I can help out.

* The text in Maps should be adjusted for the corresponding source. Perhaps:

    // The optional second return value when getting a
    // value from a map indicates if the key was present
    // in the map. This can be used to disambiguate
    // between missing keys and keys with zero values
    // like `0` or `""`. Here we didn't the value itself, so
    // we ignored it with the _blank identifier_ `_`.

* Adjust the Range source and generated output now that the blank identifier is introduced here. It may be enough just to de-italicize the "blank identifier" text in the Range page (we try to use italics when a term is first introduced).

Would you be up for taking a pass at these? Happy to help too if you'd like.

mathonsunday:
I'll take a pass at these.

mathonsunday:
My latest commit should address all your feedback.

mmcgrana:
Looks great! Thanks @mathonsunday for incorporating all the feedback so well, and again for the original suggestion.

Updated pages are live:
https://gobyexample.com/maps
https://gobyexample.com/range

Good luck with your Go work!

PR #87, opened 2015-01-08
Added helpful description for required parentheses for multiple return v...

soroushjp:
...alue functions


mmcgrana:
Thanks for the suggestion, but I'm going to try to keep this text brief for now and omit this explicit description. If others write in having run into this I may change it though.

soroushjp:
Not a worry, was only a quick suggestion, completely understand wanting to keep the text short.

- Sent from Mailbox on iPhone

PR #93, opened 2015-01-28
[examples] time-formatting-parsing: 19, 35, 48

sorinlacriteanu:
- removed error e initialization where it was not needed
- fixed last initialization


sorinlacriteanu:
Hi, I made a little improvement that might help people that actually type the examples by hand and don't run the whole thing at once. Cheers for the great examples :+1:

mmcgrana:
Thanks for the suggestion! I'd actually like to keep it the way it is though, so that it's clear from the start that the second return value is an error. Thanks again for sending in the PR though!

PR #95, opened 2015-02-12
url-parsing: Use net.SplitHostPort

philk:
Just something I stumbled across parsing an IPv6 address. SplitHostPort should work in all cases where strings.Split("url", ":") works as well as more complicated cases.


mmcgrana:
Huh, TIL! Thanks @philk.

PR #96, opened 2015-02-22
Fix comment on rand example.

skrul:
Example shows methods being called on r1 (type Rand) not s1 (type Source).  Fix comment to reflect this correctly.


mmcgrana:
Good catch, thanks!

PR #97, opened 2015-02-23
Fix a few small typos.

skrul:


skrul:
Should I be running the build and also including the generated hash and web pages in the pull request?

mmcgrana:
Thanks so much for these fixes @skrul! I can go ahead and do the rebuild after merging for a small change like this, thanks for checking though.

PR #98, opened 2015-02-25
Complete TODO for coalesced function parameters.

skrul:
You think the comment / example is sufficient?


mmcgrana:
Ah nice! This looks great.

PR #99, opened 2015-02-26
Complete type switch TODO

skrul:
Wasn't sure what level of detail was required here so kept it pretty simple.  Open to suggestions!


Keirua:
It would be great to see this one merged (after fixing the duplicate `̀the` and the weird spacing in the 2nd sentence).

mmcgrana:
This is pretty good and I would like to bring it in. Thanks for the suggestion. I will make some minor edits post-merge.

PR #100, opened 2015-02-26
Complete named return value TODO

skrul:
Included is a simple example of named returned values and a second example that shows how named return values can be useful in conjunction with defer.  The second example is pretty contrived, but I included it because it is something you can't do any other way.  However it may be too esoteric for this tutorial and I'm glad to drop it :)

### DIFF
--- a/examples/multiple-return-values/multiple-return-values.go
+++ b/examples/multiple-return-values/multiple-return-values.go
@@ -5,11 +5,44 @@
 package main
 
 import "fmt"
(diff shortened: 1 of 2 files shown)


mmcgrana:
Thanks for the suggestion, but I'm not going to expand the scope of the site to include named return values. (I think I had this in a TODO list at some point, but have decided not to expand the scope of the site in general due to personal bandwidth constraints.)

PR #105, opened 2015-04-16
Per platform 'sed' command to format files to have 4 spaces (darwin and linux supported)

rafmagana:
Based on the comments on [this commit](https://github.com/mmcgrana/gobyexample/commit/8e7a6bb08685033cddff20f626896fd269b403b7) I added conditional `sed` commands since syntax is different among OSs and this wasn't working on Linux, other people might add their own platforms by adding more cases.

### DIFF
--- a/tools/format
+++ b/tools/format
@@ -4,7 +4,24 @@ set -eo pipefail
 
 paths=$(ls tools/*.go examples/*/*.go)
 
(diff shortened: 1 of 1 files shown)


mmcgrana:
Thanks for the fix, I know this has bugged a few people!

Sorry it took so long to review this /:

PR #110, opened 2015-07-24
Update the random examples

richo:
Is the whitespace meant to be spaces here? I thought go used tabs (and evidently gofmt saw fit to tabbify) but I have no idea what I'm doing.

### DIFF
--- a/examples/random-numbers/random-numbers.go
+++ b/examples/random-numbers/random-numbers.go
@@ -4,43 +4,46 @@
 
 package main
 
+import "time"
 import "fmt"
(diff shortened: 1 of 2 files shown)


mmcgrana:
Good call to clarify this! I've edited this commit a bit and landed it here:

https://github.com/mmcgrana/gobyexample/commit/f60da41f5b74d687a480daac2212de48c9aec523

(The unusual whitespace is on purpose and due to the space and layout constraints we have in the two-column format.)

richo:
Neato, thanks!

PR #111, opened 2015-08-11
Fixed 2 typos

sbdchd:
Fixed a couple typos in the channels example.

I believe "sends and receives" should be "send and receive".


mmcgrana:
This is actually intentional, though I can se how it'd be a bit confusing. Thanks for flagging though!

PR #112, opened 2015-09-04
fix typo on random numbers example

poopoothegorilla:
fixes small typos in Random numbers example

Also is there a reason `go fmt` isn't being used in this project? If there is I can get rid of the automatic converts to tabs.


mmcgrana:
Good catch, thanks for the fix. I adjusted the format and merged this in.

`go fmt` is used (see `tools/format`) but we tweak the output slightly to accommodate the tight width restrictions on the code due to the double-column layout.

PR #115, opened 2015-11-23
Update struct example to show pointer impact

garrypolley:
When working through the struct example I think it would help if there was a final line that showed the impact of editing a struct that is referenced by a pointer.


garrypolley:
I don't think this is going to get picked up, so I'm going to close it.

PR #118, opened 2016-01-01
Added fallthrough statement to switch examples

jamesor:
I updated the switch examples to include an example of using the fallthrough statement.


mmcgrana:
Thanks for the suggestion, but I'm not planning to expand the scope of the Switch page to include `fallthrough`.

To be honest I have never used `fallthrough` in my own Go code so I'm especially hesitant to include it.

PR #120, opened 2016-01-08
slices: add an example

amitsaha:
This commit adds an example to the slices examples to demonstrate that when a slice is created from an underlying slice, any changes to the new slice will also be reflected in the original slice.


mmcgrana:
This is a good thought by I think the side effects here are a bit subtle for a beginner, so I'm not going to include it for now.

PR #122, opened 2016-01-22
Extended `for`: Added explanation of `continue`.

mhutter:


mmcgrana:
This makes sense, thanks. I will do some editing post-merge.

PR #123, opened 2016-01-25
Error checking in examples/reading-files

calmh:
This improves the error checking in in this example by clarifying that
the check() pattern should be limited to the top level of command line
programs, avoiding the full panic backtrace, and checking the return of
close() at the end. I think this better reflects current practices,
although I'm tempted to remove the check() function at all and provide
more useful error messages at each exit point - although this may make
the example too cluttered.

(I think I got it right with the formatting and hash thing, although
this was a slight fight against my editor - why not standardize on
gofmt:ed code like the rest of the Go universe? ;)

### DIFF
--- a/examples/reading-files/reading-files.go
+++ b/examples/reading-files/reading-files.go
@@ -14,9 +14,13 @@ import (
 
 // Reading files requires checking most calls for errors.
 // This helper will streamline our error checks below.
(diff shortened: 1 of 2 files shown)


calmh:
¯\_(ツ)_/¯

PR #126, opened 2016-03-07
New range example

poopoothegorilla:
I think that this example should be added... it is just a small addition, but it could provide useful to those learning.

When I ran the rebuild tool it replaced some `<em>` tags to `_` I am unsure wether this should be removed.


poopoothegorilla:
closes https://github.com/mmcgrana/gobyexample/issues/127

mmcgrana:
Per #127 I've pulled this suggestion in - thanks!

PR #134, opened 2016-05-19
Adding info on danger of leak on non-closed time Ticker.

henvic:


mmcgrana:
Thanks for the suggestion, but I'm going to keep the example as is to make it as brief as possible. Sorry!

PR #137, opened 2016-06-02
Remove unnecessary word in Range.

bitwiselove:
Fix paragraph in range.


mmcgrana:
Ah, good catch, thanks for flagging! I've fixed this in master in both the `.go` source and generated files.

PR #139, opened 2016-06-09
Add 'previous example' navigation

svanburen:


mmcgrana:
Thanks for the suggestion, but I'm not planning on changing the navigation of the site any further. Sorry!

PR #140, opened 2016-06-26
errors.go: Replace errors.New() with fmt.Errorf()

nodakai:


catsby:
This is the only line on the page that directly uses the `errors` package, so I think it's useful to leave as is, unless I'm missing something non-idiomatic or something about directly calling `errors.New` that is generally not recommended.

mmcgrana:
Yeah as @catsby mentioned I'd like to retain this as an example of using the underlying `errors` package. Thanks for the suggestion though!

PR #141, opened 2016-07-17
Add a little more detail for Go specific syntax in if statements.

katamaritaco:


Keirua:
:+1:

mmcgrana:
I'd rather not lengthen the text here, I think what's there is enough for a new user to find their way given the example code and `gofmt`.

PR #146, opened 2016-10-16
Fix typo in the command line flags example.

b5y:
This change is [Reviewable](https://reviewable.io/reviews/mmcgrana/gobyexample/146)


mmcgrana:
Thank you for flagging! Fixed in ba1fb3a which also updated the generated output.

PR #147, opened 2016-11-08
added clipboard functionality

jarnix:
Hi,

First, thank you for this tool that helped me learn go :)

I added a button to copy the code to clipboard. It's easier to play in my editor than online since I can store the changes locally and have syntax highlighting (play.golang.org is too minimalist).

Basically, there is an external js (clipboard.js hosted from a CDN) added in template/example.html, a site.js that contains the code to setup the copy functionality, the clipboard icon, and I rebuilt the files with tools/build.

There was also a typo in tools/format with sed (sed -i '' was not working in my debian, I changed to "sed -i" without the empty quotes, maybe you will want to change it back)


jarnix:
Hi @mmcgrana, sure, will try to do it asap :)

PR #148, opened 2016-11-21
Add clarification about index and length of string

Guitarbum722:
Adding a note to `string-functions.go` that explains that the index of a string is actually a byte of that string, and the length of a string is the length in bytes; not characters which would only technically work for ASCII characters in the string.  A link is provided to the applicable blog post.


mmcgrana:
Good call. I ended up editing strings-functions a bit to incorporate this idea and link.

PR #150, opened 2016-12-22
Update string-functions

stevejs:
Fix typo so that displayed output matches displayed source code.


mmcgrana:
Thanks for flagging! I've addressed this in a separate commit that also updates the corresponding .sh file.

PR #151, opened 2017-01-03
added clipboard functionality with your comments

jarnix:
Hi @mmcgrana ,

Sorry for opening a new PR, I suck at git.

What is the copyright status of the clipboard image?

-> it's a new one that I drew myself, so it's okay now

When I try this locally an extra newline gets appended to the copied text. Would it be possible to fix that? (I see that this currently works by walking through the HTML; I would also be open to just dumping the raw source code somewhere in the DOM so that it is easier to pluck out at copy-time.)

-> it's done

I just merged a change that should fix sed on Linux. Could you try out the latest master and see if that works for you? Then we should back out the sed change in this PR.

-> it's ok :)

There are a few conflicts from individual examples. It may be easiest to limit the PR to the source changes since I'm going through and fixing a bunch of small things right now. I can run build locally to see the effect of the changes.

-> also done, forgive me :)

We should probably just inline clipboard.js similar to how we do with site.css. The site is deployed on a CDN so we'd still get the performance benefit but would only depend on 1 CDN.

-> done


eliben:
FWIW this would be really nice to have. It really bothered me a couple of times in the past as a user - the only reasonable way to copy paste the sample was to click through to the source or to the playground.

mmcgrana:
OK! Finally merged this.

I tried it out and it looks great 👍 Just made some tweaks due reconcile with current master. Thanks again for submitting this.

PR #153, opened 2017-01-13
Fix some style errors

oohira:
Thanks for your good document.
I fixed some style errors.


mmcgrana:
Good calls, thank you. Incorporated along with site rebuild in #190.

PR #156, opened 2017-02-09
FIxed typo

theyoux:


mmcgrana:
Fixed in #189, thanks for flagging though!

PR #159, opened 2017-04-27
Minor consistency fix

Dmitry-Erokhin:


mmcgrana:
The use of point indexes here is intentional. I'm trying to explain open range indexes assuming that people don't know how those work yet, so don't want to explain them in terms of themselves.

PR #160, opened 2017-04-27
Change style of time.Duration definition in ticker example

wallyqs:
Thought it would be more readable this way :)
Thanks for the great resource of examples!


mmcgrana:
Addressing this in the linked PR. Thank you for flagging!

wallyqs:
awesome, thanks!

PR #163, opened 2017-05-30
Readability Point

733amir:
Add a note about readability for using the name of fields when initializing a struct.


mmcgrana:
I think the text is OK without these and am not going to merge them, but thanks for the suggestion!

PR #168, opened 2017-07-06
fixed command

atilkan:


mmcgrana:
This `time` is intentional. See the timing output a few lines down in `select.sh`.

PR #169, opened 2017-07-29
Adding explanation to the non-blocking send example

arithmetric:
Adding documentation to the non-blocking send example to explain:
- Why the send is prevented
- How to adjust the code to to allow the send


mmcgrana:
Good call. I've added this, with some editing, in #194.

PR #171, opened 2017-09-03
"t declared and not used" fix

terminalcommand:
In the latest function, the switch body (line 48) causes a compiler warning "t declared and not used" 
If I make the corresponding change from switch i := i.(type) to switch i.(type), it compiles without any issues.

### DIFF
--- a/examples/switch/switch.go
+++ b/examples/switch/switch.go
@@ -43,10 +43,9 @@ func main() {
 
     // A type `switch` compares types instead of values.  You
     // can use this to discover the the type of an interface
-    // value.  In this example, the variable `t` will have the
-    // type corresponding to its clause.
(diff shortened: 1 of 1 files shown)


mmcgrana:
How do you reproduce the "t declared and not used" warning? I don't get this warning with either of:

```console
$ go build github.com/mmcgrana/gobyexample/examples/switch
$ go run examples/switch/switch.go
```

terminalcommand:
Sorry, it was an error on my part. 

I was typing the examples by hand and changing parts of them to see what happened. 

In the switch example the default case uses the variable t. I probably typed it out without  writing the default case, so the compiler complained about t being declared and not used. 

It works on my end too, sorry for the inconvenience again.

PR #174, opened 2017-11-05
fixed typo

atilkan:
fixed typo


mmcgrana:
I can't belive this website still has typos lol. Thanks for flagging. Addressing in #189 at the source.

PR #176, opened 2017-11-18
Clarify that it's 10 goroutines instead of 10 writes

terrytangyuan:


terrytangyuan:
@mmcgrana Could you take a look? Thanks!

PR #177, opened 2017-12-01
Remove duplicate `the` in `switch` page

cttttt:
Remove double `the`


mmcgrana:
This has since been fixed in another PR but thanks for flagging!

cttttt:
👍

PR #179, opened 2018-01-04
removal of unwanted variable

bharathramh92:


mmcgrana:
Fair! Incorporating into #188.

PR #180, opened 2018-01-05
BUG FIX related to deadlock

bharathramh92:
If the job channel buffer is smaller than the actual number of jobs, then the main go routine wont be able to fill the job channel and hence a possible deadlock. It has to be done in a separate go routine.

https://play.golang.org/p/eIQi0rhlXKU illustrates the issue.

### DIFF
--- a/examples/worker-pools/worker-pools.go
+++ b/examples/worker-pools/worker-pools.go
@@ -25,8 +25,8 @@ func main() {
     // In order to use our pool of workers we need to send
     // them work and collect their results. We make 2
     // channels for this.
-    jobs := make(chan int, 100)
-    results := make(chan int, 100)
+    jobs := make(chan int, 2)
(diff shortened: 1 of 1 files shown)


examples/worker-pools/worker-pools.go
```diff
@@ -25,24 +25,28 @@ func main() {
     // In order to use our pool of workers we need to send
     // them work and collect their results. We make 2
     // channels for this.
-    jobs := make(chan int, 100)
-    results := make(chan int, 100)
+    jobs := make(chan int, 2)
```
bharathramh92: this change is to highlight the reader the reason for having a separate go routine for setting up the job queue.

mmcgrana:
I see that this can deadlock if you back yourself into that corner, but I don't think we need to address that in this minimal example.

PR #196, opened 2018-02-13
Added pdf version - disabled links for now since they are not right

cgopalan:
I love Go By Example and wanted an offline version. I was able to generate a pdf using [pdfkit](https://pypi.python.org/pypi/pdfkit) .
All I had to do was [this simple code](https://gist.github.com/cgopalan/3dd23eaddb7a54f2d16671eb8347b83e).


mmcgrana:
Thanks for sharing this! The official Go by Example site is for now designed only for the web medium (including offline via local HTML files and server). So I'm not going to merge this into the main repo at this time. But hopefully if anyone else is interested in a PDF version they find their way to your code!

cgopalan:
👍 Thanks for Go By Example!

jonaustin:
Just to put this somewhere for anyone who wants a PDF version.

It's easy to create a pdf with [httrack](https://github.com/xroche/httrack) and [calibre](https://github.com/kovidgoyal/calibre).

```
httrack https://gobyexample.com/
calibre gobyexample.com/index.html
```

And then in Calibre, right click the file and choose `Convert books`, and select PDF as the output format.

PR #198, opened 2018-03-12
Added prefixes with purpose to find output result easier.

cn007b:


cn007b:
Could anyone please advise me what to do in this case?

butuzov:
Vadim, there is no point in this, really. There are quite a lot of example in gobyexample that use this structure, and part of this examples a kept short to make resulting HTML short, so it wouldn't hurt existing design, IMHO.

mmcgrana:
Coming back to this example after not reading for a while, it is tricky to read 😅 

There is some precedent for using prefixes in cases where the output would otherwise be confusing, e.g. https://gobyexample.com/slices.

I'd probably use lower case and shorter prefixes.

Assigning myself to take care of this when I have a sec.

cn007b:
@mmcgrana I can update PR and made all prefixes in lower case, if it make sense?

eliben:
> @mmcgrana I can update PR and made all prefixes in lower case, if it make sense?

Go ahead and update the PR per Mark's suggestions

cn007b:
I had to fork repo one more time 🤦‍♂️ so now I have 1 more PR: https://github.com/mmcgrana/gobyexample/pull/391
This PR will be closed.
@mmcgrana Please review PR: https://github.com/mmcgrana/gobyexample/pull/391
and let me know in case I need to improve something 🙏

eliben:
This ended up landing in a separate PR #391

PR #199, opened 2018-04-18
Article URL updated.

butuzov:
I have update Article URL in `url-parsing` example, after noticing that old URL end-up in redirection loop with SSl Error.


butuzov:
Awesome! thank you!

PR #200, opened 2018-05-13
update constants.go

strickvl:
I'm a real newbie to Go, but my teacher was telling me the other day that Go refers to 'casting' as 'conversion' and that you actually won't find the word 'cast' anywhere in the language spec for that reason. So I thought maybe it might be an idea to make that same change in how you refer to it in gobyexample?


mmcgrana:
Good catch @strickvl!

I've gone ahead and fixed that + all the other cases + rebuilt to reflect.

PR #204, opened 2018-07-29
Fix channel-synchronization example : channel buffer

YumaInaura:
In this case, channer buffer is not neccessary
Because without channel buffering, receiver wait for a spawned groutine finish.


mmcgrana:
Yeah, I don't think there is really a problem here. The buffer reflects that it's fine for the spawned routine to finish without the spawner being ready to deal with that, but the spawner can't proceed until it does eventually handle it (via the blocking receive)

PR #206, opened 2018-10-17
Update tickers.go to not leak goroutine

tschundler:
The code in the example will leak a goroutine because ticker.Stop() does not close ticker.C.
There is no way to close ticker.C, so `for _ := range ticker` is not a generally safe example for using a ticker.

### DIFF
--- a/examples/tickers/tickers.go
+++ b/examples/tickers/tickers.go
@@ -16,9 +16,16 @@ func main() {
     // `range` builtin on the channel to iterate over
     // the values as they arrive every 500ms.
     ticker := time.NewTicker(500 * time.Millisecond)
+    done := make(chan bool)
+
     go func() {
-        for t := range ticker.C {
-            fmt.Println("Tick at", t)
+        for {
+            select {
+            case <-done:
+                return
+            case t := <-ticker.C:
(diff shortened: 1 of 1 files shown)


tschundler:
I don't think my variant is that conflated, as it mirrors a realistic real-world use case.

However, that seems like a reasonable alternative; keeping the example more focused. Maybe just update the text "Once a ticker is stopped it won’t receive any more values on its channel." to "Once a ticker is stopped it won’t receive any more values on its channel, but will remain open."

and/or add more description to the `break`:
```
if t > limit {
   ticker.Stop()
   // Stopping the ticker will not close the channel and end the loop.
   break
}
```


The choice to not close to avoid a spurious tick makes sense, but also isn't obvious, and it would help to guide others away from getting bitten by that.

eliben:
Assigning to @mmcgrana for an opinion

blixt:
As I was debugging leaky goroutines in our code base I found that someone who doesn't normally code Go had copied this example. Please merge this pull request as this bad example is bound to spread itself into production code very often in similar cases.

mmcgrana:
This is a solid update. Thanks for sharing it, and apologies for the delay in merging! 😬

PR #208, opened 2018-11-11
Update variables.go

tolidano:
Note that variable declaration can happen outside function scope and shorthand is only available inside function context.


tolidano:
this was not addressed in 4 months, so either it's covered, or unmaintained.

PR #211, opened 2018-11-29
Added warning for capital field names in structs

baileywickham:
Names without capital letters will be ignored by the JSON Marshaller. This may cause confusion when fields are not being unmarshalled correctly without a warning why.

### DIFF
--- a/examples/json/json.go
+++ b/examples/json/json.go
@@ -14,8 +14,8 @@ type response1 struct {
     Page   int
     Fruits []string
 }
-// Fields must have capital letters to be exposed to other
-// programs such as the JSON Marshaller. 
+// Only exported fields will be encoded/decoded in JSON. 
+// Fields must start with capital letters to be exported.
 type response2 struct {
     Page   int      `json:"page"`
     Fruits []string `json:"fruits"`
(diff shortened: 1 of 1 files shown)


examples/json/json.go
```diff
@@ -14,6 +14,8 @@ type response1 struct {
     Page   int
     Fruits []string
 }
+// Fields must have capital letters to be exposed to other
+// programs such as the JSON Marshaller.
```
eliben: I don't quite understand what "other programs such as the JSON Marsheller" means here. Unexported fields will not be encoded **at all**, since the JSON package doesn't have access to them via `reflect`. 

I agree it's worth mentioning this, but:

1. I'd say it straight after the first struct (why just the second?)
2. I'd say something like "only exported fields will be encoded/decoded in JSON"

mmcgrana:
I agree this is a good thing to flag, and with @eliben comments.

examples/json/json.go
baileywickham: The reason I avoided commenting after the first struct was to preserve the formatting. I thought the space adjacent to the second struct was fit for the comment because the first struct is already commented. Should I still move it to before the first struct?

I agree with 2, I cleaned up the phrasing to make it clearer.

mmcgrana:
Cool, thanks @baileywickham! I'll merge this and take care of any final tweaks needed + regeneration.

PR #212, opened 2018-11-30
Fix closing-channels example - removed buffer

rdlf0:
Fixed closing channel example to match the desired output.
Using a buffered jobs channel would produce the following output:

`sent job 1`
`sent job 2`
`sent job 3`
`sent all jobs`
`received job 1`
`received job 2`
`received job 3`
`received all jobs`

### DIFF
--- a/examples/closing-channels/closing-channels.go
+++ b/examples/closing-channels/closing-channels.go
@@ -11,7 +11,7 @@ import "fmt"
 // to a worker goroutine. When we have no more jobs for
 // the worker we'll `close` the `jobs` channel.
 func main() {
-    jobs := make(chan int, 5)
+    jobs := make(chan int)
     done := make(chan bool)
 
     // Here's the worker goroutine. It repeatedly receives
--- a/examples/closing-channels/closing-channels.go
+++ b/examples/closing-channels/closing-channels.go
@@ -11,7 +11,7 @@ import "fmt"
 // to a worker goroutine. When we have no more jobs for
 // the worker we'll `close` the `jobs` channel.
 func main() {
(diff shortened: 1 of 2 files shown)


eliben:
Fundamentally it doesn't matter whether the jobs channel is buffered or not, since the sample demonstrates closing channels and earlier samples cover buffering.

If I had to guess, the buffer was added because without it the output if funky. For example, I mostly get this on my machine (w/ the buffer removed):

```
received job 1
sent job 1
sent job 2
received job 2
received job 3
sent job 3
sent all jobs
received all jobs
```

Which looks strange, even though it's correct. The Go runtime switches goroutines at sync points like blocking on channels. If the main goroutine sends to an unbuffered channel it blocks until the worker is ready, but then lets the worker run for a bit so its "received" is printed before the "sent". This order is not guaranteed, of course, but it's certainly possible.

When the channel is unbuffered there is no blocking on `jobs <- j` so no context switch. I consistently get

```
sent job 1
sent job 2
sent job 3
sent all jobs
received job 1
received job 2
received job 3
received all jobs
```

So, if anything, I'd suggest just changing the expected output (in the `.sh` file) to the above and leaving the code alone.

rdlf0:
@eliben I guess you mean that the channel is buffered for the last example.
Otherwise yeah, I agree that the focus is not on the channel buffering, so I fixed the output only and put back the buffering.

eliben:
To be honest, I'm not sure it's worth it. Execution order isn't guaranteed anywhere, and tomorrow the scheduling algorithm can change and something else will be printed. This applies to almost all gobyexample samples that use goroutines.

PR #213, opened 2018-12-05
Change example text in variables.go

coltonhurst:
Instead of `short`, which is a common programming language keyword (representing the `short` data type in many languages), use something random like 'apple'.

When I first read this in the example, I thought it was creating a variable of type `short`. It took me a few seconds to realize that the example was showing the declaration of a string, not a short.

While it may just be me (brain fart), hopefully this change would help prevent confusion when reading about variable declarations, especially when the string's value is a common type (but a type that's not in Go).

If it's not worth it, and we leave totally as is, no worries. Just an idea.

### DIFF
--- a/examples/variables/variables.go
+++ b/examples/variables/variables.go
@@ -28,7 +28,7 @@ func main() {
 
     // The `:=` syntax is shorthand for declaring and
     // initializing a variable, e.g. for
-    // `var f string = "short"` in this case.
-    f := "short"
+    // `var f string = "apple"` in this case.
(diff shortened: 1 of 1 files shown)


mmcgrana:
Fair!

coltonhurst:
Just saw this PR was accepted... totally forgot about this! Now that I've done more with Go, I have no idea why this confused me in the first place, as the "short" was wrapped in quotes...

Either way, hopefully it helps someone else in the future I guess 😅

Thanks!

PR #216, opened 2019-02-26
update for Go 1.12

go101:


go101:
Since Go 1.12, [the print order of map entries becomes constant](https://golang.org/doc/go1.12#fmt), though their iteration order is still randomized.

examples/maps/maps.sh
```diff
@@ -6,4 +6,6 @@ v1:  7
 len: 2
 map: map[k1:7]
 prs: false
-map: map[foo:1 bar:2]
+map: map[bar:2 foo:1]
+
```
eliben: Remove two extra newlines

eliben:
> Since Go 1.12, [the print order of map entries becomes constant](https://golang.org/doc/go1.12#fmt), though their iteration order is still randomized.

You're right, I had forgotten about that. The PR LGTM with one minor comment to fix

mmcgrana:
Great thanks. I'll merge and fix + build on master.

Issue #166, opened 2017-06-18
How about mentioning waitGroup in the Worker Pool example?


amosshapira:
As far as I follow the Worker Pool example, the only reason it stops on time is because the final loop to print the results is careful to loop exactly 5 times. There is no way for the `worker` routines to close the `results` channel to mark that they are all done.

Perhaps it would be useful to mention the option of `waitGroup` (https://golang.org/pkg/sync/#WaitGroup), which is nicely demonstrated in https://nathanleclaire.com/blog/2014/02/21/how-to-wait-for-all-goroutines-to-finish-executing-before-continuing-part-two-fixing-my-ooops/

(EDIT: Point to an updated blog post with corrected examples)

mmcgrana:
@eliben wdyt?

eliben:
Sounds like a good idea. I can add it in #233 or do is separately in a follow-up

mmcgrana:
Let's add it in #233. I think it help make clear / depend on how/where the new example fits in the overall site.

eliben:
FWIW, I think the worker pool example is fine as is, and I wouldn't add a `WaitGroup` to it. *Mentioning* it in the comments may be a good idea, however.

The example is fine because it does what it intends to do in a simple way. We know how many results to expect, so that's how many results we read from a channel. Everything should exist cleanly and not leave goroutines hanging in this case. Note that the number of goroutines is different from the number of jobs - by design since this is a pool. A `WaitGroup` is more idiomatic for waiting for goroutines rather than for jobs, so I don't see a very natural way of incorporating it into the example without over-complicating it.

mmcgrana:
Right, I agree on mentioning it but not changing the worker pool example.

It be good to work in somewhere this idea of where a pool vs. a group is appropriate, or perhaps link to something that does, but that's quite subtle so I'm also OK if it's left implicit.

eliben:
I think this can be closed, now that #223 was merged

PR #223, opened 2019-05-29
Add "Waiting For Goroutines To Finish" example.

eliben:
Also link to it from the Channel Synchronization example


examples.txt
```diff
@@ -23,6 +23,7 @@ Goroutines
 Channels
 Channel Buffering
 Channel Synchronization
+Waiting For Goroutines To Finish
```
mmcgrana: If there's a word or small phrase in Go for the concept covered by an example I tend to use that, even if that word or phrase isn't intelligible on its own. E.g. "Select" instead of "Waiting on multiple channel operations". Sometimes there isn't a word in go, e.g. for "Non-Blocking Channel Operations". So here I'd favor title "WaitGroups" and slug `/waitgroups`.

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
```diff
@@ -0,0 +1,47 @@
+// To wait for multiple goroutines to finish, we can
+// use a sync.WaitGroup.
```
mmcgrana: By convention we italic important terms when we first introduce them in an example.

mmcgrana: We usually use the plain-English word(s) when introducing a concept and throughout the example when we refer to it in the text. For example in /timers we introduce "timers" and refer to the "timer" not `time.Timer`. WaitGroup is a little less obvious since it's not a standard CS phrase (I don't think?). But e.g. "we can use a _wait group_ ..." and "`wg` is the wait group it uses ..." doesn't seem obviously right either. I'm not sure.

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+    "fmt"
+    "math/rand"
+    "sync"
+    "time"
+)
+
+// This is the function we'll run in every goroutine.
+// wg is the WaitGroup it uses to notify that it's done.
```
mmcgrana: variables form the code should be monotyped: "`wg` is the WaitGroup" etc.

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
```diff
@@ -0,0 +1,47 @@
...
+// Note that a WaitGroup must be passed to functions by
+// pointer.
+func worker(id int, wg *sync.WaitGroup) {
+    fmt.Printf("Worker %d starting\n", id)
+
+    // Sleep for a random duration between 500-700 ms
```
mmcgrana: Do you feel like the randomization here is essential?

examples.txt
mmcgrana: I'm not sure where this example should go. This is one possibility. Another would be before or after the Worker Pool example (they seem related)? Indeed I'm curious to hear your thoughts on the question I linked from the corresponding issue another user filed about Worker Pools and WaitGroups.

mmcgrana:
@eliben the core of this example is solid; I think it'll prove a very helpful addition!

I've added some comments around presentation, relation to other examples, and relation to a good question in another issue. LMK what you think about those?

mmcgrana:
(I'm using assignee to indicate next action on the issue, so feel free to assign me after you respond. This just helps me differentiate which issues are on me to action vs. which I'm waiting on others for. But I don't feel strongly about this workflow, lmk if you have other preferences.)

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
eliben: Not essential, but for the presentation I think it's useful to not expect things to be deterministic. If all workers waited for time T, main could have just `time.Sleep(T+eps)` too wait for them, so the example is less convincing.

We could do something alternative like issue some HTTP HEAD requests to a few common URLs, instead.

I don't feel strongly about this, if you think we could replace the sleeps by:

```
// ... do work here
```

I'm OK with that too.

eliben:
Thanks for the review, I'll do another round and will update the PR tomorrow.

Actually the assignee workflow works perfectly for me because I can check all the issues/PRs assigned to me easily instead of having to hunt for them. So feel free to assign to me issues where you ask a question and expect my answer too

examples.txt
eliben: Alright, moving to `waitgroup`. Named it in singular to match the type, similarly to `Select` (not `Selects`), but if you feel that `WaitGroups` is better than `WaitGroup`, LMK. Also moving it to after the worker pool, which will mention it

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
eliben: Changed heading comment, refactored the other comment.

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
eliben: Removed this line since it's too obvious, IMHO and mentioned later too.

examples/waiting-for-goroutines-to-finish/waiting-for-goroutines-to-finish.go
eliben: I went ahead and changed that to just `time.Sleep(time.Second)` to match what the worker pool sample is doing. Consistency is important

eliben:
PTAL, I believe I've addressed the comments. There are some additional points to decide listed there (like the exact name). Also moved the sample to after worker pools and mentioned it there - seems like a natural location.

Fixes #166

examples/channel-synchronization/channel-synchronization.go
```diff
@@ -1,6 +1,8 @@
 // We can use channels to synchronize execution
 // across goroutines. Here's an example of using a
 // blocking receive to wait for a goroutine to finish.
+// When waiting for multiple goroutines to finish,
+// you may prefer to [use a WaitGroup](waiting-for-goroutines-to-finish).
```
mmcgrana:
```suggestion
// you may prefer to use a [WaitGroup](waiting-for-goroutines-to-finish).
```

(as you do in the worker pools example.)

examples/waitgroup/waitgroup.go
```diff
@@ -0,0 +1,42 @@
...
+// Note that a WaitGroup must be passed to functions by
+// pointer.
+func worker(id int, wg *sync.WaitGroup) {
+    fmt.Printf("Worker %d starting\n", id)
+
+    // Sleep to simulate an expensive task.
+    time.Sleep(time.Second)
+    fmt.Printf("Worker %d done\n", id)
+
+    // Notify the WaitGroup that we're done.
```
mmcgrana:
```suggestion
    // Notify the WaitGroup that this worker is done.
```

maybe? just to make it clear that the _WaitGroup_ is not necessarily done.

examples/waitgroup/waitgroup.go
```diff
@@ -0,0 +1,42 @@
...
+    wg.Done()
+}
+
+func main() {
+
+    // This WaitGroup is used to wait for all the
+    // goroutines launched here to finish.
+    var wg sync.WaitGroup
+
+    // Launch several goroutines and increment the WorkGroup
```
mmcgrana:
```suggestion
    // Launch several goroutines and increment the WaitGroup
```

I think? And one more below. Sorry I had seen this earlier but totally spaced on commenting on it.

examples.txt
mmcgrana: Yes let's use the plural. Good observation on select, bu the slightly more nuanced convention is singlural for keyword syntax (if, else, switch, select) and plural for entities (maps, arrays, tickers, worker pools).

mmcgrana:
Great! Some local comments. But otherwise I think this is ready! I'll take one more look once these final changes land but 🤞 your first example lands shortly!

eliben:
> Great! Some local comments. But otherwise I think this is ready! I'll take one more look once these final changes land but crossed_fingers your first example lands shortly!

I applied your suggestions via the Github interface - thanks!

examples/waitgroup/waitgroup.go
```diff
@@ -0,0 +1,42 @@
...
+    var wg sync.WaitGroup
+
+    // Launch several goroutines and increment the WaitGroup
+    // counter for each.
+    for i := 1; i <= 5; i++ {
+        wg.Add(1)
+        go worker(i, &wg)
+    }
+
+    // Block until the WorkGroup counter goes back to 0;
```
mmcgrana:
```suggestion
    // Block until the WaitGroup counter goes back to 0;
```

mmcgrana:
@eliben just flagging the pluralization comment above in case that got lost in the GitHub noise.

When you're ready please go ahead and assign back to me.

eliben:
> @eliben just flagging the pluralization comment above in case that got lost in the GitHub noise.
> 

Oops, yes. Will do

mmcgrana:
https://gobyexample.com/waitgroups 🎉

PR #224, opened 2019-05-29
Try fixing bios's vet

eliben:


eliben:
@nzoschke I see "BIOS Pending -- Vetting failed" above, but why did it fail? Where is the log that can help debug this?

nzoschke:
Strange. I'll look into the root cause but BIOS isn't well-supported these days. Go ahead and delete bios.sh completely and none of the check should run and fail.

mmcgrana:
I've gone ahead and removed bios.sh on master for now.

PR #225, opened 2019-05-29
Rewrite the local server in Go instead of Python

eliben:
`http.FileServer` makes writing static servers even shorter than in Python :)


mmcgrana:
Nice!

PR #226, opened 2019-05-30
Adding a basic sanity testing tool to run during 'build'.

eliben:
The tool just runs 'go vet' for now; could do more later (like actually
run the examples and compare to some expected output)


mmcgrana:
Cool 😎

PR #228, opened 2019-05-31
Basic HTTP client sample

eliben:


mmcgrana:
Great! I made a few stylistic changes. Please see commit messages for more info.

One other note: this and any other network examples we make won't run in the Go Playground. I'm fine with that, and even leaving the gopher link in place.

Please LMK if everything lgty and if you have any other thoughts on the playground link.

eliben:
LGTM!

Agree about the playground link. We *could* mention something about it in the text/comment, but I don't think that's necessary

PR #230, opened 2019-06-01
Use tabs as the canonical source indentation in git

### LINKED ISSUE #192: Try to convert to tabs in source
Then convert to spaces at build-time.

eliben:
Space conversion is done during generation only. Fixes #192


eliben:
The delta is huge because it reformats all the samples in git to use tabs. The interesting changes are in:

* `tools/format`
* `tools/generate.go`
* `tools/measure.go`

All the rest was generated when `tools/build` is run.

Note that `tools/format` stays and just runs `gofmt` on all files - I think it's good to keep it, to ensure consistent formatting of all source files on git.

mmcgrana:
Nice! Thanks.

PR #233, opened 2019-06-04
Remove pointers from the stateful-goroutines example.

### LINKED ISSUE #221: Stateful Goroutines - Why passing pointer to channel?
Hello,
First of all thanks for your website, it is useful for beginner like me :)

I wonder why do you use a pointer when you create a channel?
https://github.com/mmcgrana/gobyexample/blob/c43c319d2684b5ad7601e97b5fc2ba54552de4c3/examples/stateful-goroutines/stateful-goroutines.go#L46

I tried without passing pointer and it seemed to work : 
https://play.golang.org/p/saQTLpdIgp2

Am I wrong?

I tried to understand why it was like this and I stumbled on theses stackoverflow posts:
https://stackoverflow.com/a/44351176
https://stackoverflow.com/questions/41178729/why-passing-pointers-to-channel-is-slower

Thanks!

eliben:
The pointers are unnecessary and make the sample slightly more
complicated.

Fixes #221


examples/stateful-goroutines/stateful-goroutines.go
```diff
@@ -28,6 +28,7 @@ type readOp struct {
 	key  int
 	resp chan int
 }
+
```
mmcgrana: I had these blank line omitted so that the "These readOp and writeOp structs encapsulate..." comment clearly references both of the struct definitions, instead of just `readOp`. Do you feel strongly about adding a line here?

mmcgrana:
Thanks! Just one thing inline I wanted to get your thoughts on.

examples/stateful-goroutines/stateful-goroutines.go
eliben: Nope, no strong feelings. Wasn't aware what it was for, and the whitespace seemed a bit non-idiomatic. The formatting looks decent still:
 
KD24fCcytAb

But if prefer to keep it as before, no worries - I can roll that part back. Let me know

mmcgrana: OK yeah let's roll back to no space there. It is less idiomatic code formatting but I do prefer the resulting render.

eliben: Done

PR #234, opened 2019-06-04
Fix reading-files example to not show out-of-bounds bytes

### LINKED ISSUE #165: File read into a []byte slice issue...
https://gobyexample.com/reading-files

"Read some bytes from the beginning of the file. Allow up to 5 to be read but also note how many actually were read."

    b1 := make([]byte, 5)
    n1, err := f.Read(b1)
    check(err)
    fmt.Printf("%d bytes: %s\n", n1, string(b1))

Thing is, if less than 5 bytes are read, the remainder of the []byte slice will still contain NULL (\0) characters, and these will become part of string(b1) and be printed (perhaps invisibly, but still).

On the other hand, say the file is 6 bytes long. If the slice is reused later to read more bytes, the 2nd read will change the first byte only, and the rest of the slice will retain its contents from the first read.

If one wants to print this correctly, one should do this I guess:

    fmt.Printf("%d bytes: %s\n", n1, string(b1[0:n1]))

eliben:
Fixes #165


public/reading-files
```diff
@@ -29,7 +29,7 @@ reading files.</p>
...
+	        <a href="http://play.golang.org/p/cocJ6kBH_iZ"><img title="Run code" class="run" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAPCAYAAADtc08vAAAABGdBTUEAANbY1E9YMgAAABl0RVh0U29mdHdhcmUAQWRvYmUgSW1hZ2VSZWFkeXHJZTwAAAJ1SURBVCjPY/j//z8DMu7o6GAAgpQgD9tLqcmJH4KDg14aaik/MtdXe2ZjY6OCrh6Fs2jRYmZ9Pd05M9uL/u9dPfU/CLS0dfxvKIz/X5Dg/z8pKdkGqwGpqakMUdExDHJSYqt37tjxf+qUSf9rc2P+79298/+RA3v+H1zV///o6r7/DrbWFQkJiQwxMTGoBjAxMTpKiQmuqMuP/f/xw/v/J0+f/W9tbvTfxVLn/8rJVf+v757z/96hRf8TQtxuCQmLMjk4OKAawMfDVWVvrvd85eTq/7tXTP6/e/XM/22lif9LCnL+b13Q/v/Kzln/L++c/X/7/Jb/VpYWuZFRUagGAAErEBtlxvi+vn944f9L26cDNcz6v21R9/8zm6aC2SBDbu+f/78kK+4/L79AO7oBYCAqxD/57JZp/y/tmPX/wrYZ/6+CbAayD6zs/78daBjIgPayFJAGG6wGAIFAcpjH/dv7F4ANABuya/b/Od3l/ye2V/+/tnv2/7ldxSANmrgMYGBhZg7fuagD7GyYIeeBrrqwdRrQgLn/l02sBGkwwWkAEAjV5EZ/vQV0LswAGAYZsLC3DKTBAJ8BzCkRni/uHFyIYcAtoNc6ypL/ANVIohigrKwMxqqqqgxMzKzM6VHeL+6iGQAKzDtAV5XlJv3n5uFLRTHgzZs3YPzz50+GwqJiPitD9Y8Pjy4BB+CNvfP+3wUmIpAhhckhr3X19LodHZ28UQxQU1MDYw0NDQYBAQEeoBOTK7JjP2xf3Pt/bkfB/4KkoDcKMmIL5OXlFerq6hhu3rzJgC8MwMDYxGSfm5vbVn9/f0cgVxAkpqioyFBfX49iAACbTAK+xT3CzgAAAABJRU5ErkJggg==" /></a>
```
eliben: I wonder if the `` should be on the next line... Right now the `href` changes for every code change but the diff looks bloated because the image data is included

mmcgrana: Ha, I was just thinking about this too! I tried fiddling with the template earlier today but I was getting weird diffs on `tools/build` due to unexpected new whitespace (I think?) Anyways it'd be great to fix this.

eliben: Sent #236 to fix it. The unexpected new whitespace is due to a stray tab in the `.tmpl` file, it's now replaced by spaces for consistency with the rest of the file. I think it's just a one-off.

PR #240, opened 2019-06-05
Add VERBOSE option to tools/build and generate.go

eliben:
VERBOSE will be set by TRAVIS (but can also be set by user to diagnose slow
builds). Using an env var so that it automatically propagates to all the
sub-scripts and tools without having to pass it through tools/build explicitly.

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -2,7 +2,9 @@
 
 set -e
 
-verbose() [[ -v VERBOSE ]]
+verbose() {
+	! test -z "$VERBOSE"
+}
 
 verbose && echo "Running tests..."
 tools/test
--- a/tools/build
+++ b/tools/build
@@ -2,7 +2,16 @@
 
 set -e
 
+verbose() [[ -v VERBOSE ]]
+
+verbose && echo "Running tests..."
 tools/test
+
+verbose && echo "Formatting code..."
 tools/format
+
+verbose && echo "Measuring line lengths..."
 tools/measure
+
+verbose && echo "Generating HTML..."
 tools/generate
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,7 @@ install:
   - go get -u github.com/russross/blackfriday
 
 script:
-  - 'tools/build'
+  - tools/build
+
+env:
+  - VERBOSE=1
--- a/tools/generate.go
+++ b/tools/generate.go
@@ -285,7 +285,7 @@ func renderIndex(examples []*Example) {
 
 func renderExamples(examples []*Example) {
 	if verbose() {
-		fmt.Println("Generating HTML")
+		fmt.Println("Rendering templates")
 	}
 	exampleTmpl := template.New("example")
 	_, err := exampleTmpl.Parse(mustReadFile("templates/example.tmpl"))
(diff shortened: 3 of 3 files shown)


tools/build
```diff
@@ -2,7 +2,16 @@
 
 set -e
 
+verbose() [[ -v VERBOSE ]]
```
mmcgrana: This doesn't seem to work on my machine. I'd expect it to since I'm just on a stock Mac with stock Bash. Thoughts?

```console
$ tools/build
tools/build: line 5: conditional binary operator expected

$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

mmcgrana:
I'm :+1: on the spirit of this PR and the output in Travis lg, so I think we just need to figure out the Bash issue.

tools/build
eliben: That's odd... Please try now, I rewrote it such that it should work on OSX bash too now

mmcgrana: Ok that works for me now, thanks!

mmcgrana:
Hmm so trying this out with an empty cache now, I notice that when in `generate.go`, there is a long period of no output, followed by a flurry of all the lines for the 70 examples appearing at ~once. So I don't think we've quite achieved the goal of providing continuous visibility and helping users diagnose slow builds.

I think the issue is that we're giving example-by-example feedback in the "rendering" stage, but this stage is cheap (apply Go template + write local file). What's expensive is the "parse" stage, which is perhaps misnamed, as this is where we will a) do `pygmentize` and b) reach out to play.golang.org if there are cache misses - these are the expensive operations.

So if we want to have per-example output, can we associate it with these heave operations? Either by moving the log lines (easiest) or moving the heavier operations to the rendering functions?

Sorry I didn't catch this earlier /:

eliben:
> Hmm so trying this out with an empty cache now, I notice that when in `generate.go`, there is a long period of no output, followed by a flurry of all the lines for the 70 examples appearing at ~once. So I don't think we've quite achieved the goal of providing continuous visibility and helping users diagnose slow builds.
> 
> I think the issue is that we're giving example-by-example feedback in the "rendering" stage, but this stage is cheap (apply Go template + write local file). What's expensive is the "parse" stage, which is perhaps misnamed, as this is where we will a) do `pygmentize` and b) reach out to play.golang.org if there are cache misses - these are the expensive operations.
> 
> So if we want to have per-example output, can we associate it with these heave operations? Either by moving the log lines (easiest) or moving the heavier operations to the rendering functions?
> 
> Sorry I didn't catch this earlier /:

No worries, thanks for the catch.

I've seen this before and made a mental note to address it in a followup, but then forgot :)

I'll just add per-example status printouts for the parsing stage. Do you think I should also remove it for the "generating" stage, or leave it there? I suppose we should remove it, to lower the noise, since this stage is indeed fast.

mmcgrana:
SG, in that case I'd remove the prints from the "generating" stage since it's so fast.

eliben:
Done - PTAL. Per-file progress is printed in the slow path now

mmcgrana:
Great - thanks! Made one tweak before merging.

PR #243, opened 2019-06-08
Add directories example

eliben:
Note: there are other topics we may want to cover that may or may not belong in the same example:

1. `os.File` objects
2. Temporary files and directories
3. `filepath.Walk` for recursive traversal

I'm not sure how to incorporate (1). For (2), I think it makes sense to create a separate example because it covers both files and directories and has its own topic. For (3), it may actually make sense to just add a snippet to *this* example, because we're already building a dir hierarchy and listing contents non-recursively with `ReadDir`, so `Walk` seems to belong naturally.

WDYT?


examples/directories/directories.go
```diff
@@ -0,0 +1,77 @@
...
+	// Helper function to create a new empty file.
+	createEmptyFile := func(name string) {
+		d := []byte("")
+		check(ioutil.WriteFile(name, d, 0644))
+	}
+
+	createEmptyFile("subdir/file1")
+
+	// We can create a hierarchy of directories, including
+	// parents wiht `MkdirAll`. This is similar to the
```
mmcgrana:
```suggestion
	// parents with `MkdirAll`. This is similar to the
```

mmcgrana:
> 2. Temporary files and directories
> 3. `filepath.Walk` for recursive traversal
> 
> I'm not sure how to incorporate (1). For (2), I think it makes sense to create a separate example because it covers both files and directories and has its own topic. For (3), it may actually make sense to just add a snippet to _this_ example, because we're already building a dir hierarchy and listing contents non-recursively with `ReadDir`, so `Walk` seems to belong naturally.

Agree on both (2) and (3).

For (1), my intuition is just to cover it where needed in the context of more usage-oriented examples.

mmcgrana:
This example looks good mod the one typo - thanks!

eliben:
Done w.r.t. (3) - added it here, PTAL. Note about the implementation: I made `visit` a separate function mainly because of the width limitation, squeezing it on the same line with the call to `Walk` is challenging.

For (2) I'll create an issue to track.

mmcgrana:
Great! I've merged this one and the temporary files and directory example too, with minor cosmetic tweaks. Thanks for making these and sorry it took a while to get back to you on them!

PR #248, opened 2019-06-17
do not defer on close

Macilias:
according to https://www.joeshaw.org/dont-defer-close-on-writable-fil…es/ one should not use defer file.close(), its a bad pattern


examples/defer/defer.go
```diff
@@ -40,5 +40,9 @@ func writeFile(f *os.File) {
 
 func closeFile(f *os.File) {
 	fmt.Println("closing")
-	f.Close()
+	err := f.Close()
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "error: %v\n", err)
```
eliben: The `_,_` here doesn't seem necessary. Just call `fmt.Fprintf` without assigning return values

examples/defer/defer.go
```diff
@@ -40,5 +40,9 @@ func writeFile(f *os.File) {
 
 func closeFile(f *os.File) {
 	fmt.Println("closing")
-	f.Close()
+	err := f.Close()
+	if err != nil {
```
eliben: I would add a comment here so it'd show properly in the left pane - saying something like 

"It is important to check for errors when closing a file, even in a deferred function."

examples/defer/defer.go
Macilias: It's not necessary, but not explicitly carrying about the return values, has lead to this very "bug".
But surely it will run without it.

examples/defer/defer.go
Macilias: done

examples/defer/defer.go
```diff
@@ -40,5 +40,10 @@ func writeFile(f *os.File) {
 
 func closeFile(f *os.File) {
 	fmt.Println("closing")
-	f.Close()
+	err := f.Close()
+	// It is important to check for errors when closing a file, even in a deferred function.
```
eliben: Please run `tools/build` - there's a chance this will overflow the maximum column width and you'll have to reflow this comment.

examples/defer/defer.go
eliben: I wouldn't call it a bug, more like brevity for example. but since we have the separate `closeFile` function anyway, I feel it's OK to be more explicit here.

Regardless, `_, _ = ...` is not idiomatic in Go.

examples/defer/defer.go
Macilias: tools/build has terminated without errors

eliben:
LGTM.

Assigning @mmcgrana to review and merge once approved

mmcgrana:
Good point, thanks!

Merged with cosmetic tweak.

PR #249, opened 2019-06-17
Temporary files and directories sample

eliben:
Followup on discussion in #243

[this will have to be carefully merged once #243 lands because they are in the same location. I think this one should go right after the directories sample]


mmcgrana:
I merged this after a rebase, surprised GitHub didn't pick that up... Thanks again!

PR #252, opened 2019-06-22
Examples links accents removal

martinkunc:
Hi, this should be removing nonstandard accents from links, if they are used in example names.
For example Gösser -> gosser
What do you think would be good to have something like this in here ?


martinkunc:
Not that I found in english translation. I am currently trying to prepare a Czech translation of content, which have some. For Example I translated Hello world into "Ahoj světe".
I know this might fit in language specific version only, but I think other lang. editions possibly could take advantage of the functionality in the future. But it is not necessary on in here at all costs.

eliben:
Thanks for explaining the reason.

Personally, I'd prefer folks who run translation sites handle this - different languages may have different preferences on how to do this.

Assigning to @mmcgrana for final review.

mmcgrana:
Yes for now let's go ahead and keep these patch in any relevant translation downstreams. Looking at it I think it'll be manageable to keep rebased against the English master. Thanks for the suggestion though!

PR #255, opened 2019-07-01
When testing on Travis, verify that tools/build was run

### LINKED ISSUE #237: Set up CI pipeline to test PRs
Set up a CI pipeline like [Travis](https://travis-ci.org/) to run `tools/build` on each PR and check there are no errors.

This will catch compile errors, measure issues, generation issues. Checking for the exit code should be sufficient.

eliben:
This is done by routing the generation of HTML into a temporary directory,
and checking its diff agains the existing public/ when running tests.

Fixes #237

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ script:
 
 env:
   - VERBOSE=1
+  - TESTING=1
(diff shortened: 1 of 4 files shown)


mmcgrana:
Nice, thanks!

Issue #254, opened 2019-06-30
Allocation examples


PeterBocan:
Hey, I would like to add examples of Allocation, is that possible?

eliben:
Can you please be more specific and explain what you mean by `Allocation`? Sample code would be best

PeterBocan:
You can dynamically allocate objects 3 ways, one of which is less known, but it's all over the place in the examples. 
```go 
func NewPerson() *Person {
    return &Person{Name: "John Doe"} // This is a bit different.
}

func main() { 
   p1 := make([]int, 3) // p1 is []int
   p2 := new(int)  // p2 is *int
   p3 := NewPerson() // the same as p2, but with initialization.

  // there's no explicit way to free up memory due to Golang being
  // a GC language but to remove the last reference to object.
  p3 := nil
}
```
Couldn't be a bit helpful? 

Also, off topic, but when I run `build/tools` my templates get broken.

eliben:
I agree saying something about the `&StructType{...}` allocation would be good, perhaps in the Structs example?

I'd leave the other forms out, `make` comes up naturally in slices and maps, and `new` is fairly esoteric and rarely useful esp. for beginners

Would you like to try your hand at creating a PR for the Structs example adding the allocation somewhere?

PeterBocan:
Yes, once I get the generating working!

mmcgrana:
Just checking through issues - does the most recent merged PR along with the comments above mean we're set here?

PeterBocan:
Yes, we are, @mmcgrana

PR #257, opened 2019-07-01
Added a constructor to structs

PeterBocan:
As discussed in #254. I've added a constructor and initialization from within the constructor.

### DIFF
--- a/examples/structs/structs.go
+++ b/examples/structs/structs.go
@@ -50,5 +50,4 @@ func main() {
 	// Structs are mutable.
 	sp.age = 51
 	fmt.Println(sp.age)
-
 }
--- a/examples/structs/structs.go
+++ b/examples/structs/structs.go
@@ -12,6 +12,16 @@ type person struct {
 	age  int
 }
 
+// A de facto constructor of type `person`.
+func NewPerson(name string) *person {
+
+	// You can safely return a pointer to local variable
+	// as a local variable will survive the scope of the function.
+	p := person{name: name}
+	p.age = 42
+	return &p
+}
+
 func main() {
 
 	// This syntax creates a new struct.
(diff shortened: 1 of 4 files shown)


examples/structs/structs.go
```diff
@@ -38,4 +48,7 @@ func main() {
 	// Structs are mutable.
 	sp.age = 51
 	fmt.Println(sp.age)
+
+	// Call our constructor
```
eliben: I'd move this up after the `// An `&` prefix yields a pointer to the struct.` example, because it does exactly the same but with a function. The comment could say something like:

    // It's idiomatic to encapsulate new struct creation in constructor functions

examples/structs/structs.go
```diff
@@ -12,6 +12,16 @@ type person struct {
 	age  int
 }
 
+// A de facto constructor of type `person`.
```
eliben: Not sure what "de facto" means here. Maybe something like:

    // NewPerson constructs a new person struct with the given name

examples/structs/structs.go
```diff
@@ -12,6 +12,16 @@ type person struct {
...
+func NewPerson(name string) *person {
+
```
eliben: extra whitespace

examples/structs/structs.go
```diff
@@ -12,6 +12,16 @@ type person struct {
...
+	// You can safely return a pointer to local variable
+	// as a local variable will survive the scope of the function.
```
eliben: // You can safely return a pointer to a local variable. Go will
    // allocate such variables on the heap.

eliben:
When this is ready for another round of review, please use the "Re-request review" button or ping the PR with a message like "Please take a look"

mmcgrana:
Thank you!

PR #259, opened 2019-07-12
"Methods" example clarification

Lncn:
- I found the Methods example unclear as to what the difference was
  between value and pointer receiver types, or when you would use one
  over the other. I believe adding a third method that modifies a value
  in the rect struct helps get the 'by example' of this across better.

### DIFF
--- a/examples/methods/methods.go
+++ b/examples/methods/methods.go
@@ -6,17 +6,26 @@ import "fmt"
 
 type rect struct {
 	width, height int
+	label         string
 }
 
-// This `area` method has a _receiver type_ of `*rect`.
+// This `area` method has a _pointer receiver type_ `*rect`.
 func (r *rect) area() int {
 	return r.width * r.height
 }
 
 // Methods can be defined for either pointer or value
-// receiver types. Here's an example of a value receiver.
+// receiver types. This method `perim` of _value receiver type_
+// will pass the `rect` struct by value instead of reference for
+// each call.
 func (r rect) perim() int {
 	return 2*r.width + 2*r.height
+}
+
+// The method `setLabel` will modify the rect's (pointer receiver)
+// label value.
+func (r *rect) setLabel(label string) {
+	r.label = label
 }
 
 func main() {
@@ -34,4 +43,12 @@ func main() {
 	rp := &r
 	fmt.Println("area: ", rp.area())
 	fmt.Println("perim:", rp.perim())
+
+	// Consider what may happen if `setLabel` worked on a value
+	// receiver as opposed to a pointer receiver. Generally,
+	// pointer receivers are more efficient and more useful so
+	// are more common than value receiver methods.
+	fmt.Println("label:", r.label)
+	rp.setLabel("Rectangle")
+	fmt.Println("label:", r.label)
 }
--- a/examples/methods/methods.go
+++ b/examples/methods/methods.go
@@ -9,27 +9,27 @@ type rect struct {
 	label         string
 }
 
-// This `area` method has a _pointer receiver type_ `*rect`.
+// This `area` method has a _receiver type_ of `*rect`.
 func (r *rect) area() int {
 	return r.width * r.height
 }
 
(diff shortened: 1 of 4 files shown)


Lncn:
I don't understand the CI test here.  Looks like it failed because there's diff new generated site and the current site?

examples/methods/methods.go
```diff
@@ -6,19 +6,28 @@ import "fmt"
 
 type rect struct {
 	width, height int
+	label         string
 }
 
-// This `area` method has a _receiver type_ of `*rect`.
+// This `area` method has a _pointer receiver type_ `*rect`.
```
eliben: I don't think "pointer" adds much here, just leave it as it was

examples/methods/methods.go
```diff
@@ -6,19 +6,28 @@ import "fmt"
...
 func (r *rect) area() int {
 	return r.width * r.height
 }
 
 // Methods can be defined for either pointer or value
-// receiver types. Here's an example of a value receiver.
+// receiver types. This method `perim` of _value receiver type_
```
eliben: How about:

 ... Here's an example of a value receiver, which takes the `rect` struct by value instead of by reference.

examples/methods/methods.go
```diff
@@ -6,19 +6,28 @@ import "fmt"
...
+// will pass the `rect` struct by value instead of reference for
+// each call.
 func (r rect) perim() int {
 	return 2*r.width + 2*r.height
 }
 
+// The method `setLabel` will modify the rect's (pointer receiver)
```
eliben: Maybe

"`setLabel`  sets the label field to `label`. It has to be defined with a pointer receiver."

examples/methods/methods.go
```diff
@@ -34,4 +43,12 @@ func main() {
 	rp := &r
 	fmt.Println("area: ", rp.area())
 	fmt.Println("perim:", rp.perim())
+
+	// Consider what may happen if `setLabel` worked on a value
+	// receiver as opposed to a pointer receiver. Generally,
+	// pointer receivers are more efficient and more useful so
+	// are more common than value receiver methods.
+	fmt.Println("label:", r.label)
```
eliben: Why not initialize "label" in line 32?

examples/methods/methods.go
```diff
@@ -34,4 +43,12 @@ func main() {
 	rp := &r
 	fmt.Println("area: ", rp.area())
 	fmt.Println("perim:", rp.perim())
+
+	// Consider what may happen if `setLabel` worked on a value
```
eliben: I don't know if "Consider what may happen ..." is really useful for learners here. I'd jus say something shorter about setLabel being able to modify `r` because it takes a reference to the actual object rather than a copy (as with value receivers).

eliben:
> I don't understand the CI test here. Looks like it failed because there's diff new generated site and the current site?

Run `tools/build` locally and add the generated HTML to the PR. This CI check guarantees your changes will actually be published when pushed out

examples/methods/methods.go
Lncn: Good suggestion :)

examples/methods/methods.go
Lncn: Good point.

Also, I added the comment about pointer receiver methods being more efficient because I read that somewhere when doing 5 minutes of research after doing this exercise.  I wanted to give the reader a "if you don't know what to use or it doesn't matter, just use *pointer* receivers" fact to latch on to.  Is this generally true?

Lncn:
@eliben Thanks for the feedback.  I purposefully left out the generated HTML because of the contributing guidelines, so thanks for clarifying that in PR#260

eliben:
LGTM. Assigning to @mmcgrana for final review and merging if accepted

mmcgrana:
Thanks for suggesting the change. I definitely see where you're coming from here. I'm not going to apply this though, because I don't think explaining the principles of value vs. reference types is in scope for the site (as it's a general programming concept). Sorry!

PR #262, opened 2019-07-25
Fix typo in http servers example

julian-zucker:
It says "Here *out* simple response is …", but should say "Here *our* simple response is …".


mmcgrana:
Thanks, and sorry it took so long to merge this!

PR #266, opened 2019-08-05
use ordered list instead of unordered list to plan learning schedule better

computingfreak:
now visitors can find out how parts/sections are there in total and decide how much to allocate to each, 
so as to plan and actually complete every thing (or more stuff) in a define span of time.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -21,7 +21,7 @@
         browse the full list below.
       </p>
 
-      <ul>
+      <ol>
       
         <li><a href="hello-world">Hello World</a></li>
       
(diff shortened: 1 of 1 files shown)


eliben:
Thanks for the proposal

I don't think this is a good idea because when we add samples in the middle, the numbers of all the following ones will shift - which will cause confusion. Same for reordering samples for whatever reason.

>  how parts/sections are there in total and decide how much to allocate to each,

Surely visitors can count the number of sections roughly for this purpose, without explicit numbering

PR #268, opened 2019-08-24
Fix up language attribution for the Github UI

eliben:
Currently gobyexample is flagged as a CSS repository because of all the
generated HTML.

Following instructions at https://github.com/github/linguist#vendored-code to
ask Github to ignore files that are not actual origin code.


mmcgrana:
Nice, TIL!

eliben:
It works :)

screenshot-github com-2019 09 02-05_13_13

PR #270, opened 2019-09-05
Deterministic example for atomics.

### LINKED ISSUE #265: Consider simplifying the atomics example
#264 shows a user confusion about the output of the atomics example.

Right now, it's not clear what it demonstrates - that many goroutines can access the same counter without a race? But the output is non-deterministic, so it's hard to be convinced that there isn't a race.

Perhaps we could replace it by something deterministic - goroutines running a known number of increments. Without atomics this would produce gibberish or race errors, with atomics this will count as expected.

eliben:
Fixes #265

Changing to a deterministic example, as discussed. One slight omission compared to the previous version is that `LoadUint64` is not used now. IMHO it's a small loss because our main goal here is to mention atomics at all and point to the `atomic` package, rather than provide a comprehensive overview.

The big advantage of this example is that we get deterministic results *only* because we're using atomics. W/o atomics the results of this program will be wrong and will fail the race detector


examples/atomic-counters/atomic-counters.go
```diff
@@ -7,43 +7,40 @@
 
 package main
 
-import "fmt"
-import "time"
-import "sync/atomic"
+import (
```
mmcgrana: We should probably have a convention across the site for when we import multiple packages with a single statement, vs one package on each line.

In the original version of the site I tended to used the latter to reduce the length of examples (in terms of lines).

@eliben do you have an opinion on this?

examples/atomic-counters/atomic-counters.go
```diff
@@ -7,43 +7,40 @@
...
-	// being updated by other goroutines, we extract a
-	// copy of the current value into `opsFinal` via
-	// `LoadUint64`. As above we need to give this
-	// function the memory address `&ops` from which to
-	// fetch the value.
-	opsFinal := atomic.LoadUint64(&ops)
-	fmt.Println("ops:", opsFinal)
+	// Wait until all the goroutines are done.
+	wg.Wait()
+	fmt.Println("ops:", ops)
```
mmcgrana: Do you think there's a risk people think this means it's safe to read something like `ops` directly like this under concurrent access?

mmcgrana:
@eliben this looks great, a solid improvement! 2 quick Qs for you.

examples/atomic-counters/atomic-counters.go
eliben: Good point! I'll add a comment explaining this, which incidentally will also mention `LoadUint64`

examples/atomic-counters/atomic-counters.go
eliben: Ah, looking through the examples I think it's my fault that a number of them switched to line-per-import to this style :-/

My editor is doing this automatically for whenever there's more than a single import. In fact, this is how the official [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) tool is doing it, which is why most editor plugins do it these days.

Thinking about what's common in the Go world, I think that the `import ( ... )` style is more common for when there's more than a single import. Examples:

1. Tour of Go: https://tour.golang.org/welcome/4
2. Effective Go: https://golang.org/doc/effective_go.html
3. Go standard library https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L13
4. Go compiler https://github.com/golang/go/blob/master/src/cmd/compile/internal/ssa/block.go#L8
5. k8s: https://github.com/kubernetes/kubernetes/blob/master/pkg/auth/nodeidentifier/default.go#L19
6. Docker: https://github.com/docker/engine/blob/master/cli/required.go

etc.

But I see the point you're making about line per import being slightly (one line) shorter.

I'm somewhat in favor of using the more prevalent style, esp. given that many editors/IDEs pretty much force it these days, but am fine with whatever decision you want to make here. I'll be happy to send a PR to fix all samples to a single style, once we decide.

mmcgrana: I think consistently using `import ( ... )`  when there's more than important makes sense given the Go convention. A separate PR for that sounds good!

eliben: OK, opened #272 to track this

eliben:
Opened a separate issue to track `import (...)` consistency. This is ready for another review

mmcgrana:
Excellent, thank you!

Issue #207, opened 2018-11-10
Change the timeout example to use context


mdaffin:
There are some major flaws with the timeouts examples notably;

* It does not stop the processing of the function
* It leaks a channel that blocks forever
* It leaks a goroutine that sleeps forever waiting on the above channel

This makes it a fairly bad example to use in any form of production workload and should really be avoided in all workloads as it teaches some bad practices. It is far better to use the context package for timeouts. 

Would you be open to a pull request changing the example to use the context package instead? And using a now far more idiomatic pattern that does not suffer from the above problems?

I see far too many people linking to this example completely unaware of its flaws and thus using it in situations that just cause a large amount of problems.

eliben:
Can you please clarify where you see the leaks happening in the specific example here? Note that "channel leaks" is not really a thing - it's perfectly fine to leave channels open at the end of programs (they are not finite file descriptors).

FWIW we should also add an example of using context cancellation, but this example demonstrates an interesting use of timers with `select` in a very simple way, which I think is still useful. It is hopefully implied that basic samples are not always suitable for "production" uses; on the other hand, production-proofing every sample will make trivial examples very hard to do.

mdaffin:
>  it's perfectly fine to leave channels open at the end of programs 

The problem is when you use the example in the middle of a long running application. A channel and go routine costs a small amount of memory which over time builds up to a none trivial amount if you do not allow them to be cleaned up after you are done with them. If every program exited straight away it would not be an issue but go, and this pattern, is often used in the middle of long running servers which introduces problems over time.

Or worst, when people use this pattern to try and cancel computation, which you cannot do. With this pattern you can leak far more expensive resources then a simple channel or go routine.

We now have [context](https://golang.org/pkg/context/) which is a far better for asking gorouteines to stop then the patterns used in this example.

> It is hopefully implied that basic samples are not always suitable for "production" uses

Examples might not be production quality, which is ok, but neither is the pattern they are trying to teach - which is the problem. Examples should be easy to extend to production code, the patterns introduced here should not be used in production which makes them of no real values and, if anything, an anti-pattern.

It just ends up being a trap for newer users to fall into when they are learning go. Worst, a trap they only learn about when trying to use these patterns in production code. We have better patterns, we should be teaching them.

eliben:
@mdaffin you still haven't answered "Can you please clarify where you see the leaks happening in the specific example here?"

> Or worst, when people use this pattern to try and cancel computation, which you cannot do.

[...]

PR #271, opened 2019-09-05
Clarify use of buffered channel in the timeouts example.

eliben:
The buffered channel prevents goroutine leaks in case the
channel doesn't end up being read (as indeed happens to c1).

Updates #207


mmcgrana:
Great, thank you!

PR #276, opened 2019-09-12
New example: unit testing

eliben:
Note: I opted for using a single-file test in `package main` to make this work with the Go playground. It turns out the playground has a feature wherein a single file that has no `main` but does have `Test*` functions will be treated as a test file and run with `go test`. But the package has to be `main`. I think the example is still clear enough, and the main value here is demonstrating the table-driven style which is idiomatic Go.

### DIFF
--- a/examples.txt
+++ b/examples.txt
@@ -61,6 +61,7 @@ Line Filters
 File Paths
 Directories
 Temporary Files and Directories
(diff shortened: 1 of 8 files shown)


mmcgrana:
This is great, thanks @eliben!

PR #277, opened 2019-09-15
Structs - fix example output

danitello:
Quick fix to the ordering of the example output. It did not match the order that the code was printing in.


mmcgrana:
Thank you @danitello and @eliben!

PR #281, opened 2019-09-27
Structs: Fix output

petercunha:
The output of structs.go was out of proper order.


petercunha:
Reference: https://play.golang.org/p/00Yiw6xuICq

In the Go by Example website, `&{Jon 42}` appears at the bottom of the output when it should actually be in the middle. I fixed it :-)

eliben:
This seems to be a duplicate of #277

PR #282, opened 2019-10-03
Add clarification for else syntax

kwikiel:
Rationale: https://github.com/golang/go/issues/5440

else has to be in same line as closing parenthesis. Otherwise people coming from C/C++ would see not so obvious error. 

I've spent some time trying to debug basic fizz buzz code without knowing about it


mmcgrana:
Yes I agree with @eliben. Thank you though @kwikiel for the suggestion.

PR #283, opened 2019-10-06
Add missing word

arturcarvalho:


mmcgrana:
Thanks for the suggestion, but I'm fine without this word.

PR #284, opened 2019-10-09
Use SplitN rather than Split on os.Environ return value.

smoser:
Environment variables can contain a '=' in their value
   KEY=BAR=1 sh -c 'echo KEY VALUE is "$KEY"'

If you just use Split() then you may not actually get a 'pair'
but rather a map with length >= 3.


mmcgrana:
Hmm interesting. Thanks for pointing this out @smoser. This does seem more correct to me, but also pretty subtle. @eliben wdyt?

smoser:
I agree it's a small thing. But my googling bright me to this example.  I
used it, and then later realized the bug in the copied code. I figured I'd
save others the trouble if finding it themselves.

Thanks for your work. I've found it useful many times.

eliben:
Technically this is correct if env values contain `=`s. But this sample currently doesn't show values at all, so it's not buggy, strictly speaking. For showing the key, it's correct.

However, this makes me think we're conflating multiple things in this example here. Perhaps we should separate concerns and not parse the env values at all, just showing how to access `os.Environ`, basically doing this:

```
	for _, e := range os.Environ() {
               fmt.Println(e)
	}
```

We already demonstrate `strings.Split` elsewhere (in String Functions).

@mmcgrana if you think keeping the parsing/splitting here is useful, then switching to `SplitN` SGTM, but then the comment should be updated too.

eliben:
> One option is definitely to not show access to the actual environment variables (as suggested in the `Println` change above).
> 
> I think that that is kind of missing the point though. No one (or _very_ few) is going to want to read environment variables without looking at key _and_ value. So why not show how to do that properly?

That's a fair point

examples/environment-variables/environment-variables.go
```diff
@@ -27,7 +27,7 @@ func main() {
 	// get the key and value. Here we print all the keys.
 	fmt.Println()
 	for _, e := range os.Environ() {
-		pair := strings.Split(e, "=")
+		pair := strings.SplitN(e, "=", 2)
```
eliben: You should update the comment above as well, as it mentions `strings.Split`

smoser:
Will do

mmcgrana:
Thanks for the thoughtful discussion guys. Here's why I net out:

* I'll change to `SplitN` in the code so that copy-paste does the right thing.
* I'll update the comment to reference `SplitN`, but I'm not going to digress to explain the subtlety in the comments, as we've already covered strings functions in an earlier example

smoser:
Good enough for me. Thanks

PR #285, opened 2019-10-10
Fixed comment in tickers.go

gkotas:
Comment says `range` builtin used but `select` is actually used.


mmcgrana:
Good call, thanks! I'll re-build the HTML in a follow-up commit.

Issue #287, opened 2019-10-15
Can we print the function name along with the output for regexp?


vadasambar:
Context: https://gobyexample.com/regular-expressions
You have to go back and forth to see which `fmt.Println` is printing the output you are interested in.  
Can we print the function name along with the output?

eliben:
This is related to #198

mmcgrana:
Going to close as this is well tracked in the PR (which I'll keep open).

PR #288, opened 2019-10-15
Print the function being executed

vadasambar:
Refer #287


eliben:
This is similar to #198

mmcgrana:
Will look at these two at the same time. Sorry I keep delaying on them, there isn't a good answer so it's not fun to work on them /:

eliben:
Now that #198 has landed, it should be OK to redo this PR according to the same guidelines (short, lowercase-starting prefixes). I'll gladly review if someone takes this on

PR #289, opened 2019-10-21
set timeout for clients

sselinkurt:
Added requesting function using timeout with given url

### DIFF
--- a/examples/channels/channels-with-functions.go
+++ b/examples/channels/channels-with-functions.go
@@ -1,0 +1,35 @@
+//In this example, we give a boolean value into a channel by calculating an another channel
+//If the integer channel(the value is random) is divided into two, the value of boolean channel will be false
+//otherwise it will be true
+package main
+
+import (
+	"fmt"
+	"math/rand"
+)
+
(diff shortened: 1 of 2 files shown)


sselinkurt:
added channel examples

sselinkurt:
added channel examples

eliben:
It's not clear to me how this relates to #290 - can you close one of them and have just one PR open with all the details? Please provide extended motivation for the new example in the PR description.

sselinkurt:
In this simple channel-function example, we give a boolean value into a channel by calculating an another channel. If the integer channel(the value is random) is divided into two, the value of boolean channel will be false.

eliben:
Sorry, I don't see what this example shows that's materially different from the existing channel examples. One value is sent on a channel, and is received in another goroutine - this is what other examples demonstrate. 

>If the integer channel(the value is random) is divided into two, the value of boolean channel will be false.

The way I see it the boolean channel always gets `true` sent to it. This is received by the function which then overwrites the local variable with another value in case the integer is even. Again, I'm not sure what this is trying to demonstrate w.r.t. canonical usage of Go

PR #293, opened 2019-10-22
Ukrainian Translation URL changes

butuzov:
As gobyexample.com.ua going to be dropped soon,
translation moving to github pages hosting.


mmcgrana:
Thanks for your work on the Ukrainian translation!

Issue #294, opened 2019-10-22
sha1 hash isn't calculated properly.


butuzov:
There is an issue I found while working with the local version of `gobyexample`

### Description
`sha1` hash isn't calculated properly.

### Example (using `examples/arrays/arrays.go`)
```bash
shasum       - e71f2d3763eb2950727faa39b90cf1bc037d85fa
sha1sum      - e71f2d3763eb2950727faa39b90cf1bc037d85fa
go sha1      - e71f2d3763eb2950727faa39b90cf1bc037d85fa
generated    - 305975d13d24223181d13f042b290906d86c1a0e
```

----
`go sha1 ` generated by next go script (almost 100% copy/paste of format.go)

```go
package main

import (
	"crypto/sha1"
	"flag"
	"fmt"
	"io/ioutil"
)

func sha1Sum(s string) string {
	h := sha1.New()
	h.Write([]byte(s))
	b := h.Sum(nil)
	return fmt.Sprintf("%x", b)
}

func mustReadFile(path string) string {
	bytes, err := ioutil.ReadFile(path)
	check(err)
	return string(bytes)
}

func check(err error) {
	if err != nil {
		panic(err)
	}
}

func main() {
	file := flag.String("file", "", "File to get sha1 hash of")
	flag.Parse()

	fmt.Println(sha1Sum(mustReadFile(*file)))
}

```

butuzov:
While this isn't critical for this repository, I am using sha1 hashes to avoid additional generations, as part of the development pipeline and ci pipeline, and I believe having correct file hash is important to translators and adopters of gobyexample (in a way they find it usefull).

eliben:
It's not 100% clear to me what problem this aims to solve. Are you saying that the way gobyexample tooling calculates hashes should be compatible with other tools that compute hashing?

For the tooling this is a caching mechanism, which is an implementation detail. It could have been called "mycustomhash" instead of "sha1", because it could change without notice and we wouldn't want external tooling to depend on what the value of these hashes is - it should only make sense to gobyexamples' tooling.

What real issue is this causing you?

butuzov:
Well, I have change `build-loop` so it rebuilds only [changed files](https://github.com/butuzov/gobyexample/blob/ukrainian/tools/updated), I was using it while working on translations. I notice recently that go sources now are compliant with formatter, which is great! But hash still calculated on "spaced" version of files, and it causing continues rebuild of all files, not just once that changed.

[...]

PR #295, opened 2019-10-22
bugfix sha1 sum calculator

butuzov:
This bugfix implements correct way to calculate source sha1 hash, before it will changed,
by propagation of unchanged sources.

This commit will also include regenerated static files and *.hash files.

actual code change  can be checked in [tools/generate.go](https://github.com/mmcgrana/gobyexample/compare/master...butuzov:sha1-hashes-generation?expand=1#diff-8ade9e72f41688eadcd79c5c60297a68)


eliben:
Pending response on issue #294

mmcgrana:
O nice, this makes total sense!

The main reason I like this is that now when you click the gopher to go play.golang.org the code is correctly formatted there, and pressing "Format" is a no-op.

As @eliben mentioned in #294 you shouldn't relay on SHA1 specifically, or at least do so at your own risk.

Thanks for the suggestion @butuzov.

PR #296, opened 2019-10-25
Explain channel use in worker pools example

mozz100:
I'm fairly new to Go; channels and concurrency are two of the key concepts.  So when I read the worker pools example, I found it hard to understand where the number 100 came from, and why buffered channels were chosen.  I've tried to improve things with this PR.

It removes some "magic numbers" from the example and attempts to explain the use of buffered channels.


examples/worker-pools/worker-pools.go
```diff
@@ -26,9 +26,13 @@ func main() {
 
 	// In order to use our pool of workers we need to send
 	// them work and collect their results. We make 2
-	// channels for this.
-	jobs := make(chan int, 100)
-	results := make(chan int, 100)
+	// channels for this.  We use buffered channels
+	// because we want the workers to block until they
+	// receive on the `jobs` channel, but want to be able
```
eliben: IMHO the `numJobs` constant is fine, but I wouldn't change the comment here. There are several examples of channels before `worker-pools`, and buffering is explained there already. We assume some linear sequence of reading and can't explain every concept anew in every example.

mozz100: Alright.  I understand re: explaining everything everywhere :smile:

My feedback: I had read the other content and understood the difference between buffered and unbuffered channels, before suggesting this update.  It's not so much the difference between them, as the reasoning behind selecting to use unbuffered channels which I think newcomers would benefit from.

Before submitting, I read fairly widely on the internet and studied Chapter 7 of "Go in Action" (Kennedy et al, published by Manning).  That reference says "Create a pool of goroutines to perform work using _unbuffered_ channels."  So it took me quite a while to understand the choice here.

If after reading this, your point of view is unchanged, that's fine: I don't mind!  I'd love to know that I'm understanding things right (still a Go newbie) - and will submit another commit to revert the comment.

mozz100:
To give you the choice and to keep the commit history cleaner, I opened #298 which has only the magic-number-removal in it.  Feel free to close this one out without merging if you prefer to keep the comments in the code unchanged.

Thank you for gobyexample.  I find it incredibly helpful as I'm learning this new language.

eliben:
For context, #298 LGTM but I would close the rest of this PR.

mmcgrana:
I'm not going to change the comment here except for the magic numbers fix, but I'm not too surprised you've found it subtle / confusing @mozz100. Unfortunately queues are notoriously tricky, and there aren't always obviously right answers!

PR #297, opened 2019-10-27
Optional dark theme enabled.

vitaly-zdanevich:
To see it: enable dark mode in macOS or enable dark theme in Firefox.

### DIFF
--- a/public/site.css
+++ b/public/site.css
@@ -49,12 +49,27 @@ body {
     line-height: 20px;
     color: #252519;
 }
+    @media (prefers-color-scheme: dark) {
+        body {
+            color: gray;
+            background: #000;
+        }
+    }
 em {
     font-style: italic;
 }
 a, a:visited {
     color: #261a3b;
 }
+    @media (prefers-color-scheme: dark) {
+        a {
+            color: lightblue;
+        }
+        a:visited {
+            color: #99b;
+        }
+    }
+
 h2 {
     font-size: 32px;
     line-height: 40px;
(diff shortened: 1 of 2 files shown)


mmcgrana:
This was an interesting idea, but I tend to agree with the sentiment [here](https://github.com/golang/go/issues/34601#issuecomment-540106685) for the reasons discussed above. I understand _some_ users who have system-level dark mode enabled will want websites like Go by Example to appear in a dark background when they visit, but definitely not all will at this time. So then we'd need a toggle, which is not great, and you have to maintain two styles, etc.

mcluseau:
the vast majority of users enabling dark mode want as many dark things as possible. And it's easy to switch back. And if users massively want to chose site by site with a toogle, it's a browser issue IMO. Also, you only need to maintain the color overrides (and some images, maybe).

mcluseau:
From the latest Firefox it seems the trend goes as I suspected (we definitely don't want to implement that much logic):

6. Dark theme or light theme, your choice

Whether you prefer dark or light mode is really up to you to decide based on your environment, vision needs and mood. Toggle between either mode, or set Firefox to follow the settings for your Android, so if you switch automatically to dark mode at a certain time of day, your browser will follow along.

vitaly-zdanevich:
pkg.go.dev implement dark theme according to system preferences:

Screenshot_2021-09-07-15-48-43-457_org mozilla firefox

Screenshot_2021-09-07-15-51-34-214_org mozilla firefox

PR #299, opened 2019-10-28
Added example for fallthrough in switch

subsr97:
Added an example for using the keyword `fallthrough`.

### DIFF
--- a/examples/switch/switch.go
+++ b/examples/switch/switch.go
@@ -60,4 +60,21 @@ func main() {
 	whatAmI(true)
 	whatAmI(1)
 	whatAmI("hey")
+
+	// Every case includes a break by default.
+	// You can use the keyword 'fallthrough' to override this
+	// behaviour. In this example, we print 'FooBar' if num
+	// is even and 'Bar'  if num is odd.
+	evenOdd := func(num int) {
+		switch {
+		case num%2 == 0:
+			fmt.Printf("Foo")
+			fallthrough
+		case num%2 == 1:
+			fmt.Printf("Bar")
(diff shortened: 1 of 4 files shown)


examples/switch/switch.go
```diff
@@ -60,4 +60,21 @@ func main() {
 	whatAmI(true)
 	whatAmI(1)
 	whatAmI("hey")
+
+	// Every case includes a break by default.
+	// You can use the keyword 'fallthrough' to override this
+	// behaviour. In this example, we print 'FooBar' if num
+	// is even and 'Bar'  if num is odd.
```
eliben: I'm not sure about this because "break" presupposes C background which many readers may not possess. For example the Go Spec when describing `fallthrough` doesn't do this, but explains it from first principles.

IMHO this is too much of an arcana for gobyexample -- `fallthrough` is pretty rare in practice

subsr97: I agree with your first point.

But I had a need to use `fallthrough`.
Inspite of completing the Go Tour, I had to Google this up as it was not covered.
Ended  up finding this information in [Golang Wiki](https://github.com/golang/go/wiki/Switch#fall-through).

So, I felt this extra information might be helpful to at least a few.

mmcgrana:
Yeah I don't feel comfortable including this as I've never actually used `fallthrough` in my programming and am not sure the cost/benefit is high enough here.

I appreciate the suggestion and PR @subsr97 and understand you've had use for `fallthrough`, hopefully others will be able to find their way to the wiki as you did.

PR #301, opened 2019-11-19
Add example of using Context for http.Server

### LINKED ISSUE #273: Add example of using Context
Add an example of using the stdlib https://godoc.org/context package, perhaps for timeouts or something similar

See #207 for more info

eliben:
Contexts have several uses, so I was looking for one that's fairly idiomatic and not just synthetic - cancelling a HTTP handler in a server.

Fixes #273


mmcgrana:
Excellent! I like this example a lot.

I renamed it to `Context` since I think the highest order bit here is introducing `Context` (though I agree that an HTTP server is a good use case) and I like to keep the titles short.

I think we could try a bit more context on Context (heh) in the text, in terms of the general purpose of the feature, along the lines of the first sentence here: https://golang.org/pkg/context/. But I also think this is fine to ship as-is so I'll merge and we can add edits on top later.

eliben:
> I think we could try a bit more context on Context (heh) in the text, in terms of the general purpose of the feature, along the lines of the first sentence here: https://golang.org/pkg/context/. But I also think this is fine to ship as-is so I'll merge and we can add edits on top later.

Sure. Sent #309 as a followup, LMK if that's what you had in mind.

Unfortunately, I find contexts tricky to explain in simple terms, because they only come up in non-trivial programs. So while I don't expect folks to learn everything about contexts from gobyexample, the example should serve as a good refresher on the right syntax (mostly how to wait for cancellation correctly in a select case on `<-ctx.Done()`.

PR #303, opened 2019-12-12
Change chinese translator

gorda:
Hello,

The previous maintainers of the `Chinese` documents can no longer maintain the documents in time due to work reasons. 

Now they will be transferred to [gobyexample-cn](https://github.com/gobyexample-cn) maintenance, and the previous maintainers ([xg-wang](https://github.com/xg-wang)) are also in the organization. 

Can you merge this pr, it is a great glory to us, thank you very much.


mmcgrana:
Thanks!

Issue #302, opened 2019-12-11
time.Timer example could be improved


eliben:
The timer example starts a goroutine to wait for timer2 to expire, but the main goroutine exits much earlier - so even without the `timer2.Stop`, "Timer 2 stopped" will never be printed.

We should add a sleep at the end of `main` for enough time to actually see that the timer had time to expire had it not been stopped, and add a clarifying comment.

mmcgrana:
Fixed in b32972d.

PR #304, opened 2019-12-13
Clarify: Add time.Sleep for timers example

jan25:
Fixes https://github.com/mmcgrana/gobyexample/issues/302

Signed-off-by: Abhilash Gnan <abhilash.gnan@gmail.com>


examples/timers/timers.go
```diff
@@ -38,4 +38,7 @@ func main() {
 	if stop2 {
 		fmt.Println("Timer 2 stopped")
 	}
+
+	// Give timer2 enough time if it has to expire.
+	time.Sleep(2)
```
eliben: I suppose you meant `2 * time.Second`?

Also, the comment has to be beefed up a bit, I feel.

mmcgrana:
Good point that we needed to update this example. I ended up tweaking several aspects of the wording in b32972d.

Issue #182, opened 2018-02-01
Unhack italics around links


mmcgrana:
See 37740d69. Maybe upgrade to v2 of BlackFriday.

eliben:
I looked into upgrading to v2 of BlackFriday, but their module release of v2 is currently somewhat busted -- see https://github.com/russross/blackfriday/issues/500

Since the italics issue doesn't seem major, I'd suggest waiting until they unbreak their module stuff because at this point it makes sense to use modules and not the older packaging stuff.

PR #306, opened 2019-12-15
Remove pygments and add chroma

joshcarp:
- Removes vendor directory and uses chroma syntax highlighting instead
[github.com/alecthomas/chroma]()
- Add gomod 
- Generates files with new generate.go
- Syntax highlighting seems to be slightly different
I found this greatly simplifies the code dependencies


mmcgrana:
Interesting suggestion, thanks @Joshcarp!

I'm open to this change. To consider merging it we'd need to:

- Update README
- Update Travis script
- Fix highlighting so that it's the same as the current site (which may just be a matter of styling?)

LMK if you're interested in taking that on.

eliben:
Coincidentally, I notice this PR updates BlackFriday to 2.0, which might help resolve #182 
It was waiting for BlackFriday to start supporting go modules, and it seems like they do now

tools/generate.go
```diff
@@ -216,26 +188,48 @@ func parseSegs(sourcePath string) ([]*Seg, string) {
...
 	segs, filecontent := parseSegs(sourcePath)
-	lexer := whichLexer(sourcePath)
+	var goFile bool
 	for _, seg := range segs {
 		if seg.Docs != "" {
 			seg.DocsRendered = markdown(seg.Docs)
 		}
 		if seg.Code != "" {
-			seg.CodeRendered = cachedPygmentize(lexer, seg.Code)
+			seg.CodeRendered = cachedPygmentize(sourcePath, seg.Code)
```
eliben: Maybe this should be renamed now to avoid "pygment*" ?

tools/generate.go
```diff
@@ -216,26 +188,48 @@ func parseSegs(sourcePath string) ([]*Seg, string) {
...
 	}
 	// we are only interested in the 'go' code to pass to play.golang.org
-	if lexer != "go" {
+	if !goFile {
 		filecontent = ""
 	}
 	return segs, filecontent
 }
 
+func chromaFormat(code, fileExtension string) string {
```
eliben: It's a bit confusing that the second arg is called `fileExtension` when you're passing in the whole path. Can you please add a comment to this function and align the arg name with usage?

tools/generate.go
```diff
@@ -216,26 +188,48 @@ func parseSegs(sourcePath string) ([]*Seg, string) {
...
+	lexer := lexers.Get(fileExtension)
+	if lexer == nil {
+		lexer = lexers.Fallback
```
eliben: Is this actually needed? We're only formatting .go files and .sh files, right?

joshcarp:
Do you find it useful?
Sort of forgot about this one...

eliben:
> Do you find it useful?
> Sort of forgot about this one...

Yes, see Mark's reply in https://github.com/mmcgrana/gobyexample/pull/306#issuecomment-570804157 and also my review comments.

PR #312, opened 2020-01-07
Move wg.Done in waitgroup worker to defer

### LINKED ISSUE #310: WaitGroup example should defer wg.Done
Right now the worker function calls `wg.Done` at the end. Using `defer wg.Done` at entry would be more robust

eliben:
Fixes #310


mmcgrana:
Nice, thanks!

PR #320, opened 2020-02-01
Fix number of examples

gregimba:
Four examples are given but the comments refer to 3


mmcgrana:
Amazing to me that there are still mistakes like this in the site 😱 Thanks for catching!

Issue #325, opened 2020-02-26
Regular expressions - MustCompile


gren236:
Hello! I'm still learning, but this comment on https://gobyexample.com/regular-expressions page left me confused:

> When creating constants with regular expressions you can use the MustCompile variation of Compile. A plain Compile won’t work for constants because it has 2 return values.

You still won't be able to create a constant using this function and this pattern is used for **global variables** as mentioned in source code comment:

> It simplifies safe initialization of global variables holding compiled regular expressions.

So, is it a comment inaccuracy or am I wrong?

eliben:
Thanks for the report. You are correct, regexps cannot be constants. Would you like to send a PR to fix the comment?

gren236:
Sure! I'll do it as soon as I get to my workplace :)

mmcgrana:
Fixed in master - thanks!

PR #326, opened 2020-02-26
Fix MustCompile comment error

gren236:
This PR fixes a mistake about constants in `MustCompile` function description.
Solves #325


mmcgrana:
Thanks!

PR #329, opened 2020-03-09
Spelling

andrew-field:


andrew-field:
Also just an aside, I'm not sure what operating system/version/settings etc the deployment is using but when I run tools/build it causes every html file to change the `codeLines.push('')` line and replace every `=` with `\x3D` which fails the testing if committed. For example:

image

eliben:
> Also just an aside, I'm not sure what operating system/version/settings etc the deployment is using but when I run tools/build it causes every html file to change the `codeLines.push('')` line and replace every `=` with `\x3D` which fails the testing if committed. For example:
> 
> image

What are your `GOOS`, `GOARCH` and Go version?

andrew-field:
> 
> What are your `GOOS`, `GOARCH` and Go version?

go version go1.14 linux/amd64

eliben:
I think I can see it too when running with go1.14

This appears to be related to JS code added in #151 by @jarnix, and could be due to changes in HTML escaping in go 1.14 -- most likely the fix to https://github.com/golang/go/issues/35665 which went into 1.14

It's probably safe to just regenerate all files with this new escaping. I'd leave it to @mmcgrana to regenerate so he can ensure he can reproduce the changes with his setup. This means everyone generating the site should move to Go 1.14 now, but that's a good idea anyway.

mmcgrana:
Yeah indeed, I'm seeing the escaping change now that I've upgraded to 1.14. `master` now has this new escaping.

Thanks for the spelling fix!

PR #330, opened 2020-03-09
Possible patch/error checking

andrew-field:
There are some places where the error perhaps ought to be checked. Some changes (generate.go) are more prudent than others.


mmcgrana:
Good calls, thank you!

PR #332, opened 2020-03-20
NewPerson function should not be exported in structs examples. 

hallizh:
fix: exported func NewPerson returns unexported type *store.person, which can be annoying to use


mmcgrana:
Fixed on master!

PR #335, opened 2020-05-05
Improve Channel Direction example

ngfgrant:
Hi there,

I want to suggest that the channel directions example is expanded a little and given more meaningful names. 

So far this has been the only example I've really had to stop at and scratch my head - IMO there are too many `pings` and `pongs` and trying to follow what was going on got in the way of understanding the underlying concept (channel directions).

I've chosen `send/sendCh` and `forward/forwardCh` as a starting point but we could use `readCh` and `writeCh` for additional clarity?

Thought I would submit this for discussion.

Thanks


examples/channel-directions/channel-directions.go
```diff
@@ -7,24 +7,26 @@ package main
...
 }
 
-// The `pong` function accepts one channel for receives
-// (`pings`) and a second for sends (`pongs`).
-func pong(pings <-chan string, pongs chan<- string) {
-	msg := <-pings
-	pongs <- msg
+// The `fwd` function accepts one channel for receives
+// (`sendCh`) and a second for sends (`forwardCh`).
+func fwd(sendCh <-chan string, forwardCh chan<- string) {
```
eliben, line 20: While I agree that "pings" and "pongs" may not be optimal terminology, I can't say this is clearer. You use `sendCh` as a send-only channel in one place and as a receive-only channel in another place. This is confusing.

"Forward" channel is also confusing here, as it's not clear what it's supposed to forward.

ngfgrant: 😉given my translation is confusing is a good indication that the terms are not clear enough.

Maybe the function name should maybe be completely different to the channel name 

channel names = `sendOnly`, `receiveOnly`. 

functions = maybe just `a` and `b`? Unless there is a real underlying meaning to them?

eliben:
Assigning to @mmcgrana to decide.

Mark will know if this specific piece of code produced much confusion in the past. If not, there's an argument to keep it as is, since the new proposals don't seem to be perfect either.

ngfgrant:
Ok - actually the PR was to try and have a discussion around what would be a better example. 

I am not set on the changes I’ve suggested but my main point is if you are reading this for the first time the names get in the way of the concept.

Is there any other naming you think we could use here to help improve the reader understand channel directions?

eliben:
Closing this PR, as it does not represent an improvement to the example, IMHO. If someone has a better idea on how to make the example clearer, please open a new PR.

PR #341, opened 2020-10-08
added syntax

shariqkhan29:
adding syntax to code for easiness

### DIFF
--- a/examples/slices/slices.go
+++ b/examples/slices/slices.go
@@ -6,7 +6,13 @@ package main
 import "fmt"
 
 func main() {
-
+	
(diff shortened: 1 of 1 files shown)


eliben (commented):
I don't see the need for this, to be honest. The goal of gobyexample is not to reproduce documentation which is readily obtained -- https://golang.org/pkg/builtin/#make

... it's to provide examples of usage, which is precisely what the slices example is doing

mmcgrana:
Yes thanks for the suggestion @shariqkhan29, but I'm not going to plan to add this.

PR #342, opened 2020-10-18
Change t.Fail to t.Fatal

legonian:
Fail* continues execution, right function probably will be Fatal*.

https://golang.org/pkg/testing/#B.Fail
https://golang.org/pkg/testing/#T.Fatal


mmcgrana:
@eliben curious what your intent was here?

eliben (approved):

eliben:
> @eliben curious what your intent was here?

I probably meant `t.Fatal*` so it's a typo. Thanks for the fix, @legonian

Issue #46, opened 2013-05-06
Could you add an example for recover()?


LarryBattle:
Here's an example I wrote up for recover(). It would be nice if you included this in your examples.

```
package main

import "fmt"

func gandalf( doTalkTo bool ){
    defer catch()
    if(doTalkTo){
        // Panic stops execution of the current function and stops unwinding the stack, calling an deferred functions along the way.
        panic("You shall not pass!")    
    }
}
// catch() must be called as a deferred 
func catch(){
    // recover() regains control over the program execution when panic() is called.
    // recover() returns is the argument passed from panic()
    if r := recover(); r != nil {
        fmt.Println("Error:", r)
    } else {
        fmt.Println("No problems occurred")
    }
}
func main() {
    gandalf(true)
    gandalf(false)
    //unwinds to the top of the stack and ends the program
    panic("Where am I going?");
}

// Output:
// Error: You shall not pass!
// No problems occurred
// panic: Where am I going?
// goroutine 1 [running]:
// main.main()
```

Demo: http://play.golang.org/p/xu9Jd1YI0T

More information here:
<a href="http://golang.org/doc/effective_go.html#recover">Effective Go - The Program Language</a>

mmcgrana:
Sorry I'm only responding to this issue now /:

I'm actually not planning to add any more sections to Go by Example. `recover` is a very reasonable idea but due to personal bandwidth limitations I need to not increase the scope of the site at this point. Sorry!

kilaka:
Can we reopen this as the community have grown and we can/want to help :)

mmcgrana:
Yes we're in a better position to consider this now!

eliben:
An example for `recover` sounds like a good idea, but I'd personally prefer something simpler than the proposal in this issue.

kilaka:
I've created a PR using a fork.
Have two problems which I hope you have the knowledge to fix them quickly:
1. Some strange auto-generated chars, which changed from `\x3D` to `\u003D`, causing a lot of files to look as if they were changed, but actually were not.
2. Misalignment between explanation and code.

https://github.com/mmcgrana/gobyexample/pull/344

@eliben, @mmcgrana, can you see if it looks OK?
Feel free to change/comment.

eliben:
Thanks for discussion. I ended up adding a simplified example (see linked commit). Please open an issue/PR if you have questions or suggestions about the new example.

PR #344, opened 2021-02-15
recover example

kilaka:


examples/recover/recover.go
```diff
@@ -0,0 +1,49 @@
...
+// We want to recover if we want to handle a panic, stopping it from propagating upwards.
+
+package main
+
+import (
+	"fmt"
+)
+
+func main() {
+
```
eliben, line 11: There aren't many comments in this sample. Is this intentional?

examples/recover/recover.go
```diff
@@ -0,0 +1,49 @@
...
+	recoverFromBuiltInPanic(10)
+
+	fmt.Println()
+
+	recoverFromCustomPanic(-1)
+}
+
+func recoverFromBuiltInPanic(i int) {
```
eliben, line 19: Not sure if having two almost identical functions for builtin and custom panics is really useful here.

All things being equal, the sample should be as small as possible. The goal here is to remember (or copy-paste) the proper invocation of `recover` inside a `defer`, not explore different ways in which Go can panic.

A simple "custom" example should suffice. The panic example has a panic, the recover example should just recover from the same panic.

eliben (changes requested):

examples/recover/recover.go
kilaka: Added some comments in 47eed5ec17c1925520196d718befa82e69f5ccbc.
Coming from Java's "exceptions world", the recover seems a bit trivial, so I'm not sure what more info should I add.

examples/recover/recover.go
kilaka: Fixed as you requested in 47eed5ec17c1925520196d718befa82e69f5ccbc.
Thanks for the feedback.

kilaka: But I didn't use the same panic, as I think it's important to see that the panic might happen in a "lower" function call.
If it's critical I can change to align with the panic example. Or perhaps align the panic example?
Whatever you prefer...

kilaka:
Sorry guys but I'm still stuck with two problems which I hope you know how to fix quickly:
1. Some strange auto-generated chars, which changed from `\x3D` to `\u003D`, causing a lot of files to look as if they were changed, but actually were not. (Perhaps related to working on a Linux laptop?).
2. Misalignment between explanation/comments and code, specifically the comment above the recover.

eliben:
IMHO the example should be even simpler, something like (with appropriate comments):

```
func mayPanic() {
	panic("a problem")
}

func main() {
	defer func() {
		if r := recover(); r != nil {
			fmt.Println("Recovered. Error:\n", r)
		}
	}()

	mayPanic()
	fmt.Println("After mayPanic()")
}
```

The main idea is not necessarily to teach all the ways in which panic/recover works, but to provide a small copy-pasteable example folks can use if they don't exactly remember the syntax/mechanics. I agree that it's useful to place the potential panic in a function, but the recover can be in `main`, and there's no need for any additional logic that's not specifically required for the panic->recover chain to work.

But this is just my opinion :-) I'll leave it to @mmcgrana to decide.

eliben:
For backlinking, this PR is related to issue #46

eliben:
@kilaka do you intend to keep working on this?

kilaka:
@eliben , not soon. But I thought we're waiting for @mmcgrana 's feedback...

> @kilaka do you intend to keep working on this?

mmcgrana:
@kilaka I think your example is in the right direction. I agree the example should be as simple as possible,  @eliben's suggestion there seems plausible. I'd also want some brief comments about when/why one uses `recover`, especially since it is IIUC an unusual tool for application-level programers. See e.g. the Atomic Counter vs. Mutex vs. Stateful Goroutine pages for examples of lightweight when/why discussion. I've never used `recover` in my entire life as far as I know, so it's not easy for me to suggest particular comments, sorry 😬

eliben:
Thanks for discussion. I ended up adding a simplified example (see linked commit). Please open an issue/PR if you have questions or suggestions about the new example.

PR #345, opened 2021-02-15
update the project to work with go1.16+

hyangah:
After go1.16, go will use module mode by default, even when the repository is checked out under GOPATH or in a one-off directory. Add go.mod, go.sum to keep this repo buildable without opting out of the module mode.

> go mod init github.com/mmcgrana/gobyexample
> go mod tidy
> go mod vendor

In module mode, the 'vendor' directory is special and its contents will be actively maintained by the go command. pygments aren't the dependency the go will know about, so it will delete the contents from vendor directory. Move it to `third_party` directory now.

And, vendor the blackfriday package.

Note: the tutorial contents are not affected by the change in go1.16 because all the examples in this tutorial ask users to run the go command with the explicit list of files to be compiled (e.g. `go run hello-world.go` or `go build command-line-arguments.go`). When the source list is provided, the go command does not have to compute the build list and whether it's running in GOPATH mode or module mode becomes irrelevant.

Also add LICENSE file so that this repo shows up in pkg.go.dev, and add a link to the official go tutorial.


hyangah:
I think the failure is related to the change in go1.15,

https://golang.org/doc/go1.15
> html/template
> The package now uses Unicode escapes (\uNNNN) in all JavaScript and JSON contexts. This fixes escaping errors in application/ld+json and application/json contexts.

The current travis is using go1.14. Is updating it to use go1.15 an option?

mmcgrana:
Nice, thank you for the contribution. Some specific feedback:

- Yes go1.15 seems fine, and the escaping change also seems fine.
- `third_party` for pigements is fine for now. Hopefully we convert that to a Go dependency eventually anyways.
- Please update the README to reflect the changes. I think that just means removing the manual `go get github.com/russross/blackfriday` command?
- I don't want to make the license and documentation link changes here, please back those out.

With those changes we can merge this!

mmcgrana:
- Actually, `travis.yml` should probably be updated to remove the manual `go get` too right?

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.14
+  - 1.15
```
eliben, line 4: Since 1.16 was released already, should these bumps go directly to 1.16?

mmcgrana: That sounds good to me!

hyangah (commented):
Updated to use go1.16 in travis.
Removed the go get command from travis config and README. I couldn't remove 'install' section (and go mod download :-() because, otherwise, the default travis go dependency install script is used and tries to run `go get ./...`. :-( The go command is unhappy about files in third_party/pygments directory.

Reverted LICENSE file, but hopefully you address this so this appears in the pkg site. Currently forked versions are ranked higher and the contents aren't available from pkg.go.dev, which isn't great.

PR #346, opened 2021-02-17
Use chroma

TechplexEngineer:
Minimal amount of changes to use chroma. Significantly reduces build time.

Some of the syntax highlighting is slightly different. Most notably chroma highlights additional keywords.
| Using Pygments | Using Chroma |
| --- | --- |
| Screen Shot 2021-02-16 at 10 57 22 PM | Screen Shot 2021-02-16 at 10 57 13 PM |

To keep the diff small I have not removed pygments. 
I understand if the highlighting difference make this a change you are not interested in, just wanted to share the work.

I see in the previous pull request https://github.com/mmcgrana/gobyexample/pull/306 requests were made to:
- Update README
- Update Travis script

I'd be happy to tackle those additional items should the slight change in syntax highlighting be acceptable.


eliben:
Related to #306

eliben:
Once #345 lands, this should be rebased on top (similar changes with `go.mod`, etc.)

mmcgrana:
Awesome, I'd love to convert to Chroma, thanks for attempting this PR.

On highlighting, the default highlighting and styling we get with Chroma is not good for the site - e.g. the bright blue keywords all over the place is not a fit. But I think we could either make it the same or make it acceptable at least.

For the keyword highlighting, could you just change the CSS (in `site.css`) so that the color for those keywords is the same black as other parts of the code, as in the current site? Then are there any other differences? I didn't see them at a glance but only looked briefly. Hopefully they're also solvable with CSS changes.

If we're able to address that, then I'd like to merge this, after the all the cleanup mentioned above, like rebasing on the linked branch (being sure to build with go 1.15+), updating the READMe and Travis script where necessary, etc.

mmcgrana:
Just noting that #345 has landed.

eliben:
@TechplexEngineer do you intend to pursue this further? If not, I can try to pick up the PR

TechplexEngineer:
I've attempted to address all the feedback. Let me know if I've missed anything.

mmcgrana:
Thanks, this is very very close! Before merging I wanted to check every single example before/after to make sure nothing regressed. Doing this more thorough check I found some issues with highlighting of `.sh` code:

- There are weird red boxes in some examples. (Definitely need to fix this one)
- In one case the shell output after a blank line isn't grey like all other output. (Would be nice to fix this one)
- There isn't highlighting some keywords / types as in the old examples.  (I'm OK to lose this highlighting if it's hard to fix, if the other issues are corrected)

Some examples:

Line Filters branch vs. production:

Screen Shot 2021-02-24 at 11 08 17 AM

Screen Shot 2021-02-24 at 11 08 21 AM

Environment Variables branch vs. production:

Screen Shot 2021-02-24 at 11 09 18 AM

Screen Shot 2021-02-24 at 11 09 23 AM

Sorry I didn't catch these earlier. Mind taking a look?

TechplexEngineer:
Glad you caught those. Sorry I missed them. Will take a look

eliben:
Just a note that I went over a bunch of examples comparing old vs. new and didn't find anything beyond what Mark found. Spawning processes has a particularly egregious example of red rectangles :)

screenshot-127 0 0 1_8000-2021 02 24-18_08_22

TechplexEngineer:
I think I've fixed all the pesky red boxes and attempted to explain how the custom chroma rules work in the comments. Let me know if it isn't clear or if I missed anything.

mmcgrana:
Wow, nice work with the custom Chroma rules. And the comments are very clear. Thanks again for this great contribution, I'm thrilled to be off 🐍!

eliben:
:+1: nice work!

PR #350, opened 2021-02-20
Upgrade blackfriday to v2 and fix underscore italics

### LINKED ISSUE #182: Unhack italics around links

eliben:
Now that we have a go.mod, we can properly upgrade to blackfriday/v2

The output remains the same, but this PR brings back italic links with '_',
placing them into the link since this seems to be the more proper way.

Fixes #182


mmcgrana:
Excellent, thanks @eliben!

PR #360, opened 2021-04-14
Add Alfred 4 workflow to README.md

alfonmga:
Hi,

I created an [Alfred 4](https://www.alfredapp.com) workflow ([gobyexample-alfred-work](https://github.com/alfonmga/gobyexample-alfred-workflow)) to access to gobyexample.com content quickly. I think others may will find it useful so I would like to share it with everybody by including it in the README.

Thanks,
Alfon.


eliben:
Thanks for the contribution.

I'll admit I don't see the point in this change, having never used Alfred myself. Assigning to @mmcgrana to decide.

mmcgrana:
@alfonmga can you explain what Alfred is and what this workflow does, in 1 or 2 sentences? Sorry if it's obvious but like @eliben I don't use Alfred. Maybe it makes sense to start a section under Translations like Alternative Formats.

PR #363, opened 2021-04-20
Mobile layout

nalgeon:
First of all, thank you for the Go by Example!

The website looks great on the desktop, but is not adapted for phones. Currently pages look like this:

gbe-asis

With some minor CSS tweaks they could look like this:

gbe-tobe

I believe that the responsive layout is an improvement. Please consider accepting the PR if you agree.


nalgeon:
The build seems to fail because generated HTML differs from the original. But `CONTRIBUTING.md` says 'run tools/build locally and include the generated HTML in the PR'...

nalgeon:
Guys? @eliben, @mmcgrana?

eliben:
Sorry, I'll defer this one to @mmcgrana 
CSS is not my area of expertise.

mmcgrana:
Thanks for the suggestion and for making the effort to code up a prototype. Unfortunately I'm not going to make this change though. The site appearing like this on the phone is by design. One of the very foundational goals for the project is to always show entire working programs and to not break them up. I'm sorry since one could very disagree with that approach but that's what I've always wanted for this site. Sorry!

PR #367, opened 2021-05-11
Replaced Tick with NewTicker

Devorein:
Using `time.Tick` in vscode raises a warning `using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here (SA1015)`. This PR aims to solve that issue. Please note that I just started learning go a few days ago, so I might be making a the biggest mistake of my life here replacing it 😅. Thank you.

### DIFF
--- a/examples/rate-limiting/rate-limiting.go
+++ b/examples/rate-limiting/rate-limiting.go
@@ -4,7 +4,7 @@
 // elegantly supports rate limiting with goroutines,
 // channels, and [tickers](tickers).
 
-package main
+package examples
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


coder543:
FWIW (not that my opinion carries any weight around here), I don't think being in the `main` function is enough to demonstrate correct usage of `time.Tick`. The example starts two different `time.Tick` tickers -- one in the `main` function which is only used for the first half of the function (where a `time.Ticker` would have been stopped), and a second in a goroutine that lives until the program ends. Arguably, that second one is a more correct usage of `time.Tick`, but I'm wary of leaked goroutines such as that one... it's useful sometimes, but usually you need more control over the goroutine, including restarting it with a different tick rate.

One thing I don't like about the current example is that it's all stuffed into one function when two different approaches are being demonstrated... and I feel like those should be in separate functions to make that clearer.

Just some thoughts.

eliben:
Assigning to @mmcgrana for arbitration

eliben:
Closing the PR, as I don't see a significant problem with the current example. If someone wants to propose an alternative which uses correct and robust Go code, please open a new PR.

PR #371, opened 2021-05-26
Correction Method.go

nishadhsingh:
current doc incorrectly states that methods can only be defined on struct types

### DIFF
--- a/examples/methods/methods.go
+++ b/examples/methods/methods.go
@@ -1,4 +1,4 @@
-// Go supports _methods_ defined on struct types.
+// Go supports _methods_ defined on any coustom defined types.
 
 package main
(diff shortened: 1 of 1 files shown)


eliben:
It doesn't say methods can be defined **only** on struct types.

nishadhsingh:
> 
> 
> It doesn't say methods can be defined **only** on struct types.

sorry, my bad but it does strongly imply it and all the examples use only struct type

nishadhsingh:
Why is build failing? , i only changed a comment

eliben:
Sorry, I don't think this change is necessary. The text seems much clearer as it is, without the patch.

PR #375, opened 2021-06-09
Also mention another restriction of receive-only channels

coder543:


eliben:
IMHO the sample would do fine without this additional notice, but assigning to @mmcgrana for final review and maybe merge.

eliben:
Closing this PR. I would say the goal of gobyexample is not to get into the subtleties of the spec, but present example people can use to get started writing their code.

Issue #379, opened 2021-08-06
Proposal: Upgrade Signals example to use 1.16's signal.NotifyContext


anyu:
Hi there,

Go 1.16 introduced [signal.NotifyContext](https://pkg.go.dev/os/signal#NotifyContext), which handles graceful shutdowns with a little less boilerplate.

I think it'd be worth upgrading the example in https://gobyexample.com/signals so folks are aware of the more concise method.

Happy to open a PR if that sounds reasonable.

eliben:
Thanks for the suggestion; for reasons of Go backwards compatibility (similar to #343) I think this change should occur after Go 1.17 is officially released.

Feel free to submit a PR

eliben:
Looking at the PR, it's not clear to me what "graceful shutdowns" means in this context; could you clarify?

The existing example could be simplified to:

```
	sigs := make(chan os.Signal, 1)
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)

	fmt.Println("awaiting signal")
	<-sigs
	fmt.Println("exiting")
```

Which is, in fact, very similar to the example in the official documentation. I'm not sure why the goroutine was added in the first place (perhaps @mmcgrana will remember).

@anyu - how is rewriting with `NotifySignal` affect the shutdown here?

eliben:
Friendly ping @anyu

anyu:
Apologies for the delay!

You're right -- based on the original example (no contexts involved), my proposal isn't applicable.

I think in many real life scenarios where multiple goroutines are involved (eg. server request handling), a context is used to centralize cancelation handling across them. In which case, my PR would be a replacement for such boilerplate:
```
ctx, cncl := context.WithCancel(context.Background())
defer cncl()

sigs := make(chan os.Signal, 1)
signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)

go func() {
  select {
  case <-sigs:
	  cncl()
  case <-ctx.Done():
  }
}()
```
But I get that the examples in these tutorials are intended to be pretty tightly focused on one concept at a time, so keeping the simple example of signals without context is reasonable to me here.

eliben:
Correct, and a context example already exists here: https://gobyexample.com/context

PR #380, opened 2021-08-06
Update Signals example to use signal.NotifyContext

anyu:
Addresses https://github.com/mmcgrana/gobyexample/issues/379. 

- Requires upgrade to go1.16
- PR should only be merged after Go 1.17 is released in August 2021


eliben:
Note that this pending a resolution for the discussion in #379

PR #383, opened 2021-08-19
Fix output with real results

vitr:
Fix output according to the example executed at https://play.golang.org/p/_7wwI4PXw9o


eliben:
Did you run the actual example code? The playground link is a different sample.
The results seem up to date to me (the last line is truncated due to visual trimming on the web page).

Please reopen / create new PR if I'm missing something.

PR #391, opened 2021-09-03
Added prefixes with purpose to find output result easier.

cn007b:
This PR is continuation of original PR: https://github.com/mmcgrana/gobyexample/pull/198

### DIFF
--- a/examples/string-formatting/string-formatting.go
+++ b/examples/string-formatting/string-formatting.go
@@ -19,59 +19,59 @@ func main() {
 	// format general Go values. For example, this prints
 	// an instance of our `point` struct.
 	p := point{1, 2}
-	fmt.Printf("%v\n", p)
+	fmt.Printf("struct1: %v\n", p)
 
 	// If the value is a struct, the `%+v` variant will
 	// include the struct's field names.
-	fmt.Printf("%+v\n", p)
+	fmt.Printf("struct2: %+v\n", p)
 
 	// The `%#v` variant prints a Go syntax representation
 	// of the value, i.e. the source code snippet that
 	// would produce that value.
-	fmt.Printf("%#v\n", p)
+	fmt.Printf("struct3: %#v\n", p)
 
 	// To print the type of a value, use `%T`.
-	fmt.Printf("%T\n", p)
+	fmt.Printf("type: %T\n", p)
 
 	// Formatting booleans is straight-forward.
-	fmt.Printf("%t\n", true)
+	fmt.Printf("bool: %t\n", true)
 
 	// There are many options for formatting integers.
 	// Use `%d` for standard, base-10 formatting.
-	fmt.Printf("%d\n", 123)
+	fmt.Printf("int1: %d\n", 123)
 
 	// This prints a binary representation.
-	fmt.Printf("%b\n", 14)
+	fmt.Printf("bin: %b\n", 14)
 
 	// This prints the character corresponding to the
 	// given integer.
-	fmt.Printf("%c\n", 33)
+	fmt.Printf("int2: %c\n", 33)
 
 	// `%x` provides hex encoding.
-	fmt.Printf("%x\n", 456)
+	fmt.Printf("hex: %x\n", 456)
 
 	// There are also several formatting options for
 	// floats. For basic decimal formatting use `%f`.
-	fmt.Printf("%f\n", 78.9)
+	fmt.Printf("float1: %f\n", 78.9)
 
 	// `%e` and `%E` format the float in (slightly
 	// different versions of) scientific notation.
-	fmt.Printf("%e\n", 123400000.0)
-	fmt.Printf("%E\n", 123400000.0)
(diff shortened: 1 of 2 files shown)


examples/string-formatting/string-formatting.go
```diff
@@ -19,92 +19,92 @@ func main() {
...
+	fmt.Printf("int1: %d\n", 123)
 
 	// This prints a binary representation.
-	fmt.Printf("%b\n", 14)
+	fmt.Printf("bin: %b\n", 14)
 
 	// This prints the character corresponding to the
 	// given integer.
-	fmt.Printf("%c\n", 33)
+	fmt.Printf("int2: %c\n", 33)
```
eliben, line 48: I'd name this `char`, not `int2`

examples/string-formatting/string-formatting.go
```diff
@@ -19,92 +19,92 @@ func main() {
...
+	fmt.Printf("pointer: %p\n", &p)
 
 	// When formatting numbers you will often want to
 	// control the width and precision of the resulting
 	// figure. To specify the width of an integer, use a
 	// number after the `%` in the verb. By default the
 	// result will be right-justified and padded with
 	// spaces.
-	fmt.Printf("|%6d|%6d|\n", 12, 345)
+	fmt.Printf("Int3: |%6d|%6d|\n", 12, 345)
```
eliben, line 82: In these examples the salient point is the width control, not the type of the variable. I'd call these something like width{1,2,3}

Then since `int1` remains the only `int`, just call it `int`

examples/string-formatting/string-formatting.go
```diff
@@ -19,92 +19,92 @@ func main() {
...
 
 	// To left-justify, use the `-` flag.
-	fmt.Printf("|%-6.2f|%-6.2f|\n", 1.2, 3.45)
+	fmt.Printf("float5: |%-6.2f|%-6.2f|\n", 1.2, 3.45)
 
 	// You may also want to control width when formatting
 	// strings, especially to ensure that they align in
 	// table-like output. For basic right-justified width.
-	fmt.Printf("|%6s|%6s|\n", "foo", "b")
+	fmt.Printf("str4: |%6s|%6s|\n", "foo", "b")
```
eliben, line 96: These should also be `widthN`

In general, it's best not to have non-consecutive examples named `prefixN`

examples/string-formatting/string-formatting.go
```diff
@@ -19,92 +19,92 @@ func main() {
...
 
 	// To left-justify use the `-` flag as with numbers.
-	fmt.Printf("|%-6s|%-6s|\n", "foo", "b")
+	fmt.Printf("str5: |%-6s|%-6s|\n", "foo", "b")
 
 	// So far we've seen `Printf`, which prints the
 	// formatted string to `os.Stdout`. `Sprintf` formats
 	// and returns a string without printing it anywhere.
-	s := fmt.Sprintf("a %s", "string")
+	s := fmt.Sprintf("str6: a %s", "string")
```
eliben, line 104: Call this one `sprintf`

eliben (changes requested):
Thank you. I left some suggestions

examples/string-formatting/string-formatting.go
cn007b: Maybe not `width{1,2,3}` but `width`, WDYT?

examples/string-formatting/string-formatting.go
cn007b: doneYeah, agree. Done.

examples/string-formatting/string-formatting.go
cn007b: Done.

examples/string-formatting/string-formatting.go
cn007b: I've made couple rearranges and now prefixes looks easier to read and understand, and looks like now we don't have to use `width` (I like it but it take some time to understand what it means).
Please let me know WDYT?

examples/string-formatting/string-formatting.go
cn007b: Completely agree about  non-consecutive examples.

cn007b:
@eliben Thank you for review, I've made all improvements, please let me know in case I need to improve something.

eliben (changes requested):
Could you please restore the examples to their original order? The change in this PR should only be to the prefixes

examples/string-formatting/string-formatting.go
cn007b: Done.

cn007b:
@eliben Yes sure, done.
I was thinking maybe it's possible to improve non-consecutive examples as well in this PR... 🤔

cn007b:
@eliben I'd run `tools/build` and committed result just now, so I've created one more PR: https://github.com/mmcgrana/gobyexample/pull/393

PR #393, opened 2021-09-08
Updated HTML files after adding string formatting prefixes.

cn007b:


cn007b:
@eliben Just found that you already did it.

PR #400, opened 2021-11-19
don't embed the mutex

natefinch:
Embedding a mutex is a really bad idea. It exposes the Lock and Unlock methods on the enclosing type, which then allows code from outside your type to mess with your mutex.

There was no reason to embed the mutex in this example. The comment say it is idiomatic to do so, but that is untrue.  Using a mutex called "mu" (unexported) is idiomatic.

### DIFF
--- a/examples/mutexes/mutexes.go
+++ b/examples/mutexes/mutexes.go
@@ -12,29 +12,28 @@ import (
 
 // Container holds a map of counters; since we want to
 // update it concurrently from multiple goroutines, we
-// add a `Mutex` to synchronize access. The mutex is
-// _embedded_ in this `struct`; this is idiomatic in Go.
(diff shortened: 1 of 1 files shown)


eliben:
I feel your PR description is somewhat harshly worded. Embedding `sync.Mutex` is fairly common in the standard library, for example, and it's also mentioned in several presentations by the Go team over the years.

That said, I do agree that mixing up the two concepts here may not be the best idea; the goal of this sample is to present how to use a mutex, not to get into a stylistic debate about when one should embed it vs. not. I will think about this a bit more.

eliben (approved):
Thanks, I like that this is simpler. I have added an example on embedding now, so will keep it out of here.

Will apply fixes separately (generate the html, etc)

Issue #162, opened 2017-05-24
test and benchmark examples?


qighliu29:
need examples with "testing" package

mmcgrana:
I agree these are interesting packages! But I am not planning to add more sections at this time, per https://github.com/mmcgrana/gobyexample/blob/master/CONTRIBUTING.md.

PR #403, opened 2021-11-29
feat: Add benchmark example

jidicula:
Adds a simple benchmark example and its invocation.

Related to #162


examples/testing/main_test.sh
```diff
@@ -16,3 +16,29 @@ $ go test -v
     --- PASS: TestIntMinTableDriven/-1,0 (0.00s)
 PASS
 ok  	examples/testing	0.023s
+
+# Run all benchmarks in the current project in verbose
+# mode. All tests are run prior to benchmarks. The `bench`
+# flag receives a regex for benchmark function names.
+$ go test -v -bench=.
+=== RUN   TestIntMinBasic
```
eliben, line 24: I would leave the `-v` out of the benchmark example - it just repeats lots of long text

examples/testing/main_test.go
```diff
@@ -66,3 +66,12 @@ func TestIntMinTableDriven(t *testing.T) {
 		})
 	}
 }
+
+// Benchmark tests typically go in `_test.go` files and begin with `Benchmark`.
+// Each benchmark function is run several times, with `b.N` increasing on each
+// run until the `testing` runner deems the benchmark to be accurate.
+func BenchmarkIntMin(b *testing.B) {
```
eliben, line 73: Hmm, I would hope for a clearer explanation of `b.N`. What the reader needs to know is that they have to run their "thing" in a loop `b.N` times

eliben (changes requested):
Thanks, I think this is a useful contribution.

Unsure about the phrasing in the comment - open for discussion

examples/testing/main_test.sh
jidicula: Resolved in bf78b89

examples/testing/main_test.go
jidicula: Hopefully addressed in bf78b89.

examples/testing/main_test.go
```diff
@@ -66,3 +66,15 @@ func TestIntMinTableDriven(t *testing.T) {
...
 	}
 }
+
+// Benchmark tests typically go in `_test.go` files and are
+// named beginning with `Benchmark`. The `testing` runner
+// executes each benchmark function several times, increasing
+// `b.N` on each run until it collects a precise measurement.
+func BenchmarkIntMin(b *testing.B) {
+	// To measure a benchmark, the `testing` runner calls the
+	// function in a loop `b.N` times.
```
eliben, line 76: Hmm, this doesn't sound right either, sorry. The runner doesn't call our function in a loop - it's our code that does it.

I would say something like: "Typically the benchmark runs a function we're benchmarking in a loop `b.N` times"

eliben (changes requested):

eliben:
Please resolve conflicts

eliben (approved):

PR #405, opened 2021-12-16
fix: fix incorrect result output of closing-channels example

nicholasxjy:


eliben:
Please see https://github.com/mmcgrana/gobyexample#im-getting-output-in-a-different-order-from-the-example-is-the-example-wrong

PR #406, opened 2021-12-30
Correct inaccuracy in `const` description.

jginsburgn:


eliben (commented):
I've given it some thought and won't be merging this fix.

The current text is simpler while not misleading the reader, IMHO

PR #407, opened 2021-12-31
Use the full power of closure.

varunarang:
Use the full power of closure by accepting the parameters to anonymous function being run by goroutine.


eliben (commented):
The linked FAQ section (which this commit removes for some reason) mentions both approaches

IMHO the current version is fine

PR #408, opened 2022-01-03
fix: Wrong results on Closing-Channels

sang-w0o:
- Closing-Channels section seems to have wrong result when the code shown is executed.


eliben (commented):
This is not wrong; see https://github.com/mmcgrana/gobyexample#faq

PR #409, opened 2022-01-04
Update `Epoch` example to add UnixMilli from Time package

sandeepkru:
Standard library time package has support for UnixMilli - https://pkg.go.dev/time#Time.UnixMilli
Updated the example to use the method on Time instead of calculating.


eliben (approved):
Thanks! `UnixMilli` was added recently in 1.17 and I agree it's useful to include.

I'll merge this and will reword a bit in a followup

Issue #412, opened 2022-02-07
Add templates examples


ArashSameni:
can i add some examples of templates ?
something like this
https://docs.google.com/document/d/17-eD5SO8ChKi4a4DXJq24SxOgb8AYdBeEpW9pcqj1Ok/edit?usp=sharing

PR #413, opened 2022-02-08
Templates: added comments, output and listed in examples

ArashSameni:


eliben:
To link the issue and PR, this is addressing #412

eliben (changes requested):
Please remove the leftover files in `public` that remained from before the renaming of the example.

Also, if you've addressed a review comment - add an answer or just mark it "resolved".

Also - please ping the PR explicitly (or use the "ask for review" feature on the top right of the PR UI) when it's ready for another round

PR #414, opened 2022-02-09
Fix typo in strings-and-runes

kovshenin:


eliben (changes requested):
Note that you're changing the generated HTML here, whereas you want to change the source of the sample. Please see CONTRIBUTING.md for more details.

kovshenin:
Oh, I didn't realize the HTML came from the comments in the .go files, sorry about that.

eliben (approved):
Thanks for noticing this and sending a fix

PR #418, opened 2022-03-29
Update SHA1 example to SHA256

haydentherapper:
This encourages the use of stronger cryptographic hashes. While SHA1 is
still used in some applications, it's considered insecure and no longer supported
by web browsers for TLS certificates. Tools such as securego/gosec will show
errors when SHA1/MD5 hashes are used.


examples/sha256-hashes/sha256-hashes.sh
```diff
@@ -0,0 +1,15 @@
+# Running the program computes the hash and prints it in
+# a human-readable hex format.
+$ go run sha256-hashes.go
+sha256 this string
+1af1dfa857bf1d8814fe1af8983c18080019922e557f15a8a...
```
haydentherapper, line 5: `./tools/measure` won't let this line be longer than 54 characters, so I truncated the hash. Would you prefer I update `measure` to allow the full hash?

examples/sha256-hashes/sha256-hashes.go
```diff
@@ -1,26 +1,25 @@
...
 	"fmt"
 )
 
 func main() {
-	s := "sha1 this string"
+	s := "sha256 this string"
 
-	// The pattern for generating a hash is `sha1.New()`,
-	// `sha1.Write(bytes)`, then `sha1.Sum([]byte{})`.
+	// The pattern for generating a hash is `sha256.New()`,
```
eliben, line 19: Can you remove this comment entirely while you're at it? It seems wrong (the `Sum` call doesn't match the docs or the code) too

eliben (changes requested):
Thanks for the PR! This looks pretty good overall

examples/sha256-hashes/sha256-hashes.sh
eliben: Hmm, yes, it needs the size restriction to display it properly. Your workaround LGTM

haydentherapper: Sounds good, thanks!

examples/sha256-hashes/sha256-hashes.go
haydentherapper: Removed the comment.

eliben (approved):

PR #420, opened 2022-04-12
Fix a typo in the embedding example

smottt:
A small typo fix.

`Go support embedding` -> `Go supports embedding`


eliben (changes requested):
Would you like to update the source file? `public/embedding` is generated

Issue #422, opened 2022-05-13
Add embed example


peterzhu1992:
Hi,

Currently the `Embedding` example on the website is about using struct/interface to embed.
This is not all the same as `embed` introduced in 1.16 (I think)?

https://pkg.go.dev/embed
```
//go:embed
```

I am willing to help adding this example, but I want to know if this is needed, as embed is more of a directive.

Thanks.

peterzhu1992:
Sure @eliben I will try to write an example.
Tho my question is what is the second hash in the `.hash` file?
The 1st one is the sha1 value, what is the second one?

Thanks.

eliben:
> Sure @eliben I will try to write an example. Tho my question is what is the second hash in the `.hash` file? The 1st one is the sha1 value, what is the second one?
> 
> Thanks.

It's a URL hash but you should not worry about it -- just run `tools/build` to regenerate everything, it automatically creates the `.hash` files, you never have to worry about them manually.

peterzhu1992:
Got it, I didnt realize I need to update the examples.txt for it to take effect, thanks.
On another note, let me know if these are all the deliverables for a PR adding examples:

1. Entries in example.txt
2. New+modified html in public folder
3. New example/<> folder

I assume that is all I need to PR, not else as far as I can tell.

Also, do you want me to put both `struct-embedding` name change and `go:embed` in one PR or you prefer two?
I am ok with either approach tho.

Thanks.

PR #423, opened 2022-05-17
Add embed-directive example

peterzhu1992:
Signed-off-by: peterzhu1992 <peterzhu1992@gmail.com>

### Description
Add embed-directive example and rename embedding to struct-embedding.

### Resolved Issue
#422


peterzhu1992:
@eliben opened a PR, please review, please raise comments if you want me to improve anything.
It is my 1st time contribute to this repo so things can be messy. 😄 

Thanks.

eliben:
Sorry, I was meaning to comment this on the issue - let's separate the renaming into its own PR first. Once that one lands, this PR can be rebased and reviewed.

peterzhu1992:
@eliben sure, I have put this PR back to WIP.
Please approve and merge this renaming 1st.
https://github.com/mmcgrana/gobyexample/pull/424

I will rebase this PR once that one is merge.

Thanks.

peterzhu1992:
Hi @eliben,

Thanks for merging #424 and I have also resolved the conflict of this one.

Please review, thanks.

examples.txt
```diff
@@ -20,6 +20,7 @@ Structs
 Methods
 Interfaces
 Struct Embedding
+Embed Directive
```
eliben, line 23: I wouldn't put it here. After "Temporary files ...." seems like a better option

examples/embed-directive/embed-directive.go
```diff
@@ -0,0 +1,51 @@
+// Starting in Go version 1.16+, Go supports use of `//go:embed`
+// which is a directive to embed files and folders into
+// the application binary by the compiler.
+
+// Since Go does not have preprocessor or macros like C type of languages,
+// `pragmas/directives` are implemented by the compiler as comments `//`.
+// Using `//go:embed` tells the compiler to include arbitrary files/folders in binary.
```
eliben, line 7: Comment lines are too long here -- please fold all to be 80 cols at most

examples/embed-directive/embed-directive.go
```diff
@@ -0,0 +1,51 @@
+// Starting in Go version 1.16+, Go supports use of `//go:embed`
+// which is a directive to embed files and folders into
+// the application binary by the compiler.
+
+// Since Go does not have preprocessor or macros like C type of languages,
```
eliben, line 5: I'll make this all somewhat briefer, providing only the necessary information for the example at hand, w/o the background: no need to mention Go version, or the preprocessor vs. pragmas, etc.

examples/embed-directive/embed-directive.go
```diff
@@ -0,0 +1,51 @@
...
+// `pragmas/directives` are implemented by the compiler as comments `//`.
+// Using `//go:embed` tells the compiler to include arbitrary files/folders in binary.
+package main
+
+// If you are not using any exported identifiers from `embed` package directly,
+// you can add an underscore `_` before the package name, ask Go to import `embed`
+// without directly using it, like this: `_ "embed"`. In this example, we remove the
```
eliben, line 12: We don't "remove the underscore"... we just import it because we use the name

examples/embed-directive/embed-directive.go
```diff
@@ -0,0 +1,51 @@
...
+//go:embed embed-directive.sh
+var single_file_string string
+
+// Here is a similar example but include single file as `[]byte`.
+// Note that `//go:embed` can only be used in package level,
+// define it in any function will result in error.
+//go:embed embed-directive.sh
+var single_file_byte []byte
+
+// We can also embed multiple files or even folders with wildcard.
```
eliben, line 29: It's a clever idea to embed the .sh and .hash files, but it could be confusing because the readers of gobyexample are not aware of these files. Also, will this actually work on the playground? Each example has a generated playground link -- did you try to run the one generated for this new example?

I'm not sure how to overcome this...

eliben (changes requested):
Thanks for the contribution. We'll need to figure some things out before this can land

examples/embed-directive/embed-directive.go
peterzhu1992: It will not be able to run on the playground just like the reading file example on your website.

peterzhu1992: Unless go playground gives us any existing files that we can access this will not run on go playground, as embed requires pre-existing files to be presented before you can compile.

examples.txt
peterzhu1992: Valid callouts, will change.

examples/embed-directive/embed-directive.go
peterzhu1992: Will simplify.

examples/embed-directive/embed-directive.go
peterzhu1992: I was trying to make it clear to people that is not familiar with go directive, and would mistakenly think the `//` comment before `go:embed` is not needed. Will make it simpler tho and do you think providing a link on this topic would be better?

examples/embed-directive/embed-directive.go
peterzhu1992: Valid callouts.

examples/embed-directive/embed-directive.go
peterzhu1992: I can however in the `.sh` file do some creation like
```
echo "hello go" > example_single_file.txt
mkdir -p example_folder
echo "123" > example_folder/example1.hash
echo "456" > example_folder/example1.hash

go run embed-directive.go
```

Thoughts?

examples/embed-directive/embed-directive.go
eliben: Sure, just make it very brief (and a link SGTM). Like, "//go:embed is a compiler directive that ..."

examples/embed-directive/embed-directive.go
eliben: Ah good point about the reading files example.

Yes, having it laid out in the `.sh` makes more sense to me - good idea! I mostly feel uncomfortable exposing the "implementation detail" of these `.hash` files to users.

peterzhu1992: Will make some changes on that.
Thanks.

examples.txt
peterzhu1992: moved now.

examples/embed-directive/embed-directive.go
peterzhu1992: Tweaked now.

examples/embed-directive/embed-directive.go
peterzhu1992: Changed sentence.

peterzhu1992:
@eliben there are some issues with using `example_folder`.

The reason being that `//go:embed` requires the files to be presented when compiling.
And the existing measure will panic if `sourcePath` is a dir.

* If I dont include the `example_folder` and its files, error out:
```
Running tests...
examples/embed-directive/embed-directive.go:25:12: pattern example_folder/*.hash: no matching files found
```

* If I include the `example_folder` and its files, error out:
```
Measuring line lengths...
measure: examples/embed-directive/embed-directive.go:38
panic: read examples/embed-directive/example_folder: is a directory

goroutine 1 [running]:
main.check(...)
	/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/gobyexample_peterzhu1992/tools/measure.go:14
main.readLines({0xc0001507e0?, 0x0?})
	/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/gobyexample_peterzhu1992/tools/measure.go:20 +0x67
main.main()
	/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/gobyexample_peterzhu1992/tools/measure.go:32 +0xa8
exit status 2
```

It seems like we need to add a check in `measure.go` and `generate.go` to ensure readline does not happen on directories.
```
fileInfo, _ := os.Stat(sourcePath)
                if !fileInfo.IsDir() {}
```

Let me know how you think about this, I cannot think of any better solutions now.

PS: I now remember this is the reason I used `embad-directive.hash` as sample file to start with.

Thanks.

eliben:
Yes, I think it should be fine to skip directories in measure/build.

peterzhu1992:
Sure @eliben I am a bit busy these few days, will find a time slot to correct the scripts and push again.
Thanks.

peterzhu1992:
It is actually easier than I thought, so complete the changes right now.
Please review @eliben .
Thanks.

peterzhu1992:
Hi @eliben I read your follow up tweaks they are truly better.
Happy to see it shows up on the main website now.

Thanks.

PR #428, opened 2022-06-06
Fixing typo

deining:
This PR fixes a typo in `examples/xml/xml.go`.
It also updates the dependencies to their latest versions.


eliben (changes requested):
Please keep the two separate. The typo is OK, but I don't see much point in updating dependencies now. At the very least it calls for a justification

PR #429, opened 2022-06-09
Add u.Hostname() (Go 1.8+) to url.Parse example

gwillem:
https://pkg.go.dev/net/url#URL.Hostname

Thanks for making this site! You actually rank higher than the official go docs at a Google search for "golang url parse"

### DIFF
--- a/examples/url-parsing/url-parsing.go
+++ b/examples/url-parsing/url-parsing.go
@@ -34,9 +34,12 @@ func main() {
 	fmt.Println(p)
 
 	// The `Host` contains both the hostname and the port,
-	// if present. Use `SplitHostPort` to extract them.
+	// if present. Use `Hostname()` to get just the hostname. 
+	// Use `SplitHostPort` to extract host and port (fails 
+	// if port is missing).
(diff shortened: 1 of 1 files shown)


examples/url-parsing/url-parsing.go
```diff
@@ -34,9 +34,12 @@ func main() {
 	fmt.Println(p)
 
 	// The `Host` contains both the hostname and the port,
-	// if present. Use `SplitHostPort` to extract them.
+	// if present. Use `Hostname()` to get just the hostname. 
+	// Use `SplitHostPort` to extract host and port (fails 
+	// if port is missing).
 	fmt.Println(u.Host)
-	host, port, _ := net.SplitHostPort(u.Host)
+	fmt.Println(u.Hostname()
```
eliben, line 41: Did you try to run this sample? I'm not sure it compiles

eliben (commented):
Thanks for the contribution.

This is something I'd like to think about. I'm not sure how many of these functions/methods we want to enumerate without over-burdening the example, which is already rather long.

PR #432, opened 2022-07-09
Fix typo in Println statement

mcncl:
The example on [Generics](https://gobyexample.com/generics) shows 
```go 
fmt.Println("keys m:", MapKeys(m))
```

but the expected output shows
```go
keys: [4 1 2]
list: [10 13 23]
```

In order to align the example output for `keys:` with that of `list:` the `m` must be removed from the code.


eliben (changes requested):
Good catch, thank you!

Could you please run `tools/build` and add the generated HTML file in `public` to the PR as well? This will help the CI pass cleanly

mcncl:
@eliben thanks for the feedback! I've generated those files and updated this PR.

eliben (approved):
Thanks!

PR #433, opened 2022-07-10
Remove extra white space in switch example

ahbanavi:
As the example results, the white spaces are redundant.
https://github.com/mmcgrana/gobyexample/blob/0eaaca29eff7930d5bd8264e054637afcb835351/examples/switch/switch.sh#L2


eliben (commented):
I'm not sure why you're saying these spaces are redundant - can you provide a playground example?

PR #434, opened 2022-07-11
Examples: fix invalid html

deining:
The [W3C Validator service](https://validator.w3.org/) reveals that all examples of the gobyexample site contain invalid html:

See e.g. the [go routines example](https://validator.w3.org/nu/?showsource=yes&doc=https%3A%2F%2Fgobyexample.com%2Fgoroutines)

This PR fixes this. It also adds a `lang` attribute to the html tag, as advised by the validator.

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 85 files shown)


eliben:
The same services gives me >200 errors on HN and same for reddit.com (just two sites I randomly checked).

Have to think about it a bit, as I'm not sure I see the point in this.

PR #437, opened 2022-08-21
Update generics.sh

pavelgit:
the keys of the map
`var m = map[int]string{1: "2", 2: "4", 4: "8"}`


eliben (commented):
Go map iteration order is randomized

PR #441, opened 2022-09-12
fix(): grammar correction on select page

grejojoby:
grammar correction in statement in select page.


eliben (changes requested):
Sorry, this fix makes no sense to me. It turns a perfectly reasonable sentence into a non-grammatical one

PR #444, opened 2022-09-23
Update recover.sh

Naoya9922:
Thanks for the awesome work!
Would it be nice if, like the other examples, the command example was shown in recover.sh?


eliben (approved):
Thanks for the contribution!
I'll merge this and will re-generate the output on my end

PR #445, opened 2022-09-26
Update generics.go

orhan-erday:
I just remove the make function from the MapKeys method, the new version looks cleaner.

### DIFF
--- a/examples/generics/generics.go
+++ b/examples/generics/generics.go
@@ -14,7 +14,7 @@ import "fmt"
 // `V` has the `any` constraint, meaning that it's not
 // restricted in any way (`any` is an alias for `interface{}`).
 func MapKeys[K comparable, V any](m map[K]V) []K {
-	r := make([]K, 0, len(m))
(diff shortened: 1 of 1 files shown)


eliben (commented):
Indeed, just a `var` is simpler. However the `make` is there for a reason. We know the size of the result slice, so allocating it with a known capacity ahead of time is efficient since it avoids spurious allocations (and later GC). I feel like it's important to show how to write efficient code, especially in generic utilities that may end up being used extensively.

Maybe there should be a comment about this....

orhan-erday:
> Indeed, just a `var` is simpler. However the `make` is there for a reason. We know the size of the result slice, so allocating it with a known capacity ahead of time is efficient since it avoids spurious allocations (and later GC). I feel like it's important to show how to write efficient code, especially in generic utilities that may end up being used extensively.
> 
> Maybe there should be a comment about this....

You're right, thanks for the info @eliben, I'm going to close the PR.

PR #451, opened 2022-10-19
Update range-over-channels.go

QGerrit:
Would have saved me a fair bit of time understanding channels, since the existing code doesn't make it obvious that `<-` is blocking to someone new to Go.


QGerrit:
If I'd actually read things in order, rather than jumping ahead to what I thought I wanted, I would have known this already... maybe not worth adding this here.

PR #453, opened 2022-11-27
Update comment in Atomic Counter

Py-Himanshu-Patel:
Type: Refactoring
Adding the same text message from `atomic-counters.sh` file in examples to the comment in the code file `atomic-counters.go`.


eliben (commented):
Sorry, I don't see the point of this change

Py-Himanshu-Patel:
No worries, I will try harder next time. :sweat_smile:

Issue #438, opened 2022-08-29
Proposal: Add a little more info about short variable declaration


slab713:
I think it would be helpful if there'd be another sentence about how short variable declaration with `:=` is not possible outside of a function (including `main()`).

https://gobyexample.com/variables
https://go.dev/tour/basics/10

eliben:
Feel free to propose a PR with the change

grejojoby:
Have done the suggested changes in PR: #455 .

PR #455, opened 2022-12-14
Added more information about shorthand variable declaration

grejojoby:
Have added information on shorthand variable declaration as per requested in the following issue.
Issue: #438


eliben (commented):
I'm not sure why you've chosen to place this note in the `.sh` section? I would think a shorter note in the text itself would be more useful - e.g. "this syntax is only available inside functions".

PR #460, opened 2022-12-24
Add toggle theme button to the footer

vicradon:


eliben (changes requested):
Your PR doesn't provide any context / background on what it's doing.

vicradon:
Hi @eliben 

The PR adds a button to the footer that can toggle the site's theme between dark and light mode.

Here's the light mode view
light-mode-toggle-theme

Here's the dark-mode view
dark-mode-toggle-theme

eliben:
Thank you for clarifying.

At this time we don't be adding this feature. Dark mode can be accomplished by means of browser extensions very easily for people to whom this matters; gobyexample has a very bare-bones HTML and CSS that makes it particularly simple.

We don't want to be maintaining the extra complexity in the frontend code for something few users ever requested.

ragavendra:
> Thank you for clarifying.
> 
> At this time we won't be adding this feature. Dark mode can be accomplished by means of browser extensions very easily for people to whom this matters; gobyexample has a very bare-bones HTML and CSS that makes it particularly simple.
> 
> We don't want to be maintaining the extra complexity in the frontend code for something few users ever requested.

At least in my opinion it is nice to have it as most websites have it these days and presumably this is one of the first page a developer willing to learn Go can land? I did not use Mozilla's mode settings until for this. If the button is not possible, a note mentioning to use the browser's theme button should help IMO.

PR #463, opened 2023-02-28
creating a fuzzing example using a calculator.

Philip-21:
I created a Fuzz example here you can look it up and see if there are changes to be made. The Fuzz function accepts inputs in binary and converts them into strings and integers then executes the inputs when the calculator function is called  .
I also want to ask how the output will be generated in the .sh file.


eliben (changes requested):
Sorry I don't understand this example.

For fuzz testing you're supposed to write a `Fuzz*(t *testing.F)` function invoked with `go test -fuzz`. I don't see anything of the sort here.

Philip-21:
Ok i would make changes to that.

Philip-21:
@eliben 
In my second commit , i readjusted the code and created a test for each of the functions
The code i created here is a simple calculator .
The first function is a calculator function
 The second function sends input to the calculator function ,then  calls and executes the calculator function to display the result of the inputs.
Then for the tests 
func calculator(num1, num2 float64, operator string) having a test case of  	
func FuzzTestCalculator(f *testing.F)
and 
func InputCalculator(data float64) having a test case of 
func FuzzTestInput(f *testing.F)
I ran all tests and they Passed

examples/fuzzing/calculator_test.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	case "*":
+		result = num1 * num2
+	}
+
+	return result, nil
+}
+
+/*Accepts inputs, calls the calculator function and executes inputs and return results*/
+func InputCalculator(data float64) float64 {
```
eliben, line 39: Not sure what's going on here. `InputCalculator` does not appear to be using the `data` parameter at all - what's doing the fuzzing here?

eliben (commented):
I think it would help if you spend some time learning how fuzz tests in Go work. Here are some good resources to review:

* https://go.dev/doc/tutorial/fuzz
* https://go.dev/security/fuzz/

Philip-21:
> 

Thanks i  will through the docs , i just browsed through the docs though and from what i have seen  I should use the parameters in the  InputCalculator  function so the fuzz test can be performed appropriately when its called . I will take time to read the docs and work on it

examples/fuzzing/calculator_test.go
```diff
@@ -1,79 +1,88 @@
...
+	//convert to float
+	var floatList []float64
+	for _, b := range bytesList {
+		f, err := strconv.ParseFloat(string(b), 64)
+		if err != nil {
+			return 0, err
+		}
+		floatList = append(floatList, f)
+	}
+
```
Philip-21, line 37: Based on the code i just pushed  , i created and utilized the []byte parameter which i converted to a []float to perform the calculations .

examples/fuzzing/calculator_test.go
```diff
@@ -1,79 +1,88 @@
...
-	})
-
-}
+	//convert to bytes
+	var b []byte
+	for _, n := range testcases {
+		b = strconv.AppendFloat(b, n, 'f', -1, 64)
+		b = append(b, ',')
+	}
+	f.Add(b) //seed corpus inputs performed
```
Philip-21, line 84: i created testcases in []float and converted them to []bytes
then Added some seeds inputs into the fuzzCalculator(f*testing.F)

examples/fuzzing/calculator_test.go
```diff
@@ -1,79 +1,88 @@
...
-		_, _ = calculator(num1, num2, operator)
+	//Runs the fuzz function for testing
+	f.Fuzz(func(t *testing.T, numbers []byte) {
+		_, err := Calculator(numbers)
+		if err != nil {
+			return
+		}
+		t.Log("Calculations Test Passed")
 	})
 }
```
Philip-21, line 94: Runs the FuzzCalculator() for  testing

examples/fuzzing/calculator_test.go
Philip-21, line 84: i created testcases in []float and converted them to []bytes
then Added some seeds inputs into the fuzzCalculator(f*testing.F)

examples/fuzzing/calculator_test.go
Philip-21, line 94: Runs the FuzzCalculator() for testing

examples/fuzzing/calculator_test.go
```diff
@@ -1,79 +1,88 @@
...
 		}
+		result = floatList[0] / floatList[1]
+
 	case "*":
-		result = num1 * num2
+		result = floatList[0] * floatList[1]
 	}
 
 	return result, nil
 }
```
Philip-21, line 73: Based on the code i just pushed , i created and utilized the []byte parameter which i converted to a []float to perform the calculations . And i Performed the Calculations based on the indices of the FloatList

examples/fuzzing/calculator_test.go
Philip-21, line 84: i created testcases in []float and converted them to []bytes
then Added some seed inputs into the fuzzCalculator(f*testing.F)

examples/fuzzing/calculator_test.go
Philip-21, line 94: Runs the FuzzCalculator() for testing . The test case passed when i tested the code

Philip-21:
@eliben is the code okay . Or should changes be made

eliben:
> @eliben is the code okay . Or should changes be made

No, IMHO it's not OK. The example is too complicated, and I don't think it's useful. The official go.dev tutorial at https://go.dev/doc/tutorial/fuzz has a much simpler, more comprehensible example, so the way things stand now I don't think this is useful. I also don't think this PR review is going anywhere, as there have been multiple rounds already. Perhaps fuzzing is just not a great match for gobyexample.

Issue #421, opened 2022-04-21
Add fuzzing example


andofcourse:
Huge fan of Go By Example, I've found it more helpful than the official go docs sometimes :)

I was wondering if there were any plans to add a fuzzing example now that [Go 1.18 supports native fuzzing](https://go.dev/blog/go1.18)? According to the most recent dev survey, [most devs aren't familiar with fuzzing](https://go.dev/blog/survey2021-results#fuzzing), so I imagine an example of how it works would be a good way to get them started.

We are happy to contribute as well if you're swamped with other work. Let me know if there's a specific process we have to follow or if we should just open a PR

eliben:
Agree that having a simple fuzzing example would be beneficial. PRs welcome

kderck:
This would be great.

Philip-21:
I would like to contribute on this aspect

Philip-21:
Just created a Pr on this

eliben:
Creating a simple and useful example for fuzzing turns out to be nontrivial. https://github.com/mmcgrana/gobyexample/pull/463 one example of an attempt that didn't go well.

PR #474, opened 2023-05-02
Introduce anonymous structs

niyumard:


examples/structs/structs.go
```diff
@@ -50,4 +50,15 @@ func main() {
 	// Structs are mutable.
 	sp.age = 51
 	fmt.Println(sp.age)
+
+	// Structs don't always have to be defined as instances of types,
```
eliben, line 54: Thanks for the contribution, it's useful and I'm inclined to accept it.

Let's work out the details: I feel this comment is inaccurate because `dog` is in fact an instance of a type, just a type that isn't named. It should say something more to the tune of unnamed struct type (but also the word `anonymous` should be there.

niyumard: Sorry for the late response. 
Which one do you think is better?

> Structs don't always have to be defined as instances of named types, you can define a struct anonymously.

or

> Structs can be instances of unnamed types, therefore you can define structs anonymously.

or we could simply go with "You can define structs anonymously".

eliben (approved):
I'll approve this for now and will tweak the comment a bit

niyumard:
> I'll approve this for now and will tweak the comment a bit
@eliben
Thanks! But I'd argue against mentioning "used for a single value" because it could be misleading. My main use of anonymous structs is for table driven tests where I can do something like [this](https://quii.gitbook.io/learn-go-with-tests/go-fundamentals/roman-numerals#refactor-1).
Maybe table driven tests should go hand in hand with #421 and have their own section in the site though.

niyumard:
Looks great, thank you.

PR #475, opened 2023-05-07
update slice & add radix tree

jiangbo0216:
this is my first contribution to "gobyExample" 
please help me done this pull request

I have some question
1. Is there any command can auto update *.sh file Or we can only update it manually
2. maybe We can clarify the steps for the pull request. or please give the link for it if already have the steps 

this pull request give more use case for slice & give a simple implementation for radix tree

please help sincerely


eliben (changes requested):
The slices example should have additional comments to explain what is being added.

I'm not sure the radix tree example is in line with what gobyexample tries to show - I'd skip it

eliben:
Thanks for the contribution - I left comments. As for your questions:

1) No, it's manual, because it needs to intersperse comments there too
2) See https://github.com/mmcgrana/gobyexample/blob/master/CONTRIBUTING.md

eliben:
> > The slices example should have additional comments to explain what is being added.
> > I'm not sure the radix tree example is in line with what gobyexample tries to show - I'd skip it
> 
> Thanks for your reply
> 
> I think gobyexample mostly pay attention to the basic golang features. if that is right? I will skip it too.
> 
> I wanna to know if we can keep it? Maybe one day we wanna add some basic algorithm?

No, I don't think the radix tree example is appropriate for gobyexample

Would you like to continue the discussion on the proposed change to the slices example, or would you prefer to close the PR entirely?

jiangbo0216:
> > > The slices example should have additional comments to explain what is being added.
> > > I'm not sure the radix tree example is in line with what gobyexample tries to show - I'd skip it
> > 
> > 
> > Thanks for your reply
> > I think gobyexample mostly pay attention to the basic golang features. if that is right? I will skip it too.
> > I wanna to know if we can keep it? Maybe one day we wanna add some basic algorithm?
> 
> No, I don't think the radix tree example is appropriate for gobyexample
> 
> Would you like to continue the discussion on the proposed change to the slices example, or would you prefer to close the PR entirely?

Thanks for your kindness.  I have removed the radix tree code and add comment for slice code. Please take a look and let me know your thoughts.

examples/slices/slices.go
```diff
@@ -5,15 +5,32 @@ package main
 
 import "fmt"
 
+func handleIndexOutOfBoundsPanic() {
```
eliben, line 8: I prefer to leave this function out of the example. The example doesn't have any other functions, and this is not necessary

eliben (approved):
Thanks -- I will merge this and then do some fixups afterwards

PR #477, opened 2023-05-25
Small grammar correction

mnm364:
"it’s possible to pass a capacity explicitly as an additional parameter ot make"

"it’s possible to pass a capacity explicitly as an additional parameter **to** make"


eliben:
Thanks. Can you run `tools/build` and update the PR? This should make the test pass

Issue #417, opened 2022-03-20
Add some example for new operator `~`


0xff-dev:
First of all, thank you very much for providing so many examples. `Go1.18` released the generic, learned the example of the generic from the website of the repository, and found that there was a lack of introduction to the `~` operator.  Should we add an example of the operator `~`?

eliben:
Open to suggestions here. A good, useful example can be added to the Generics page -- make suggestions in a PR

PR #482, opened 2023-08-22
add an example for operator '~'

0xff-dev:
add an example for operator '~'

Fix: https://github.com/mmcgrana/gobyexample/issues/417

The following is a preview of the effect:

image

image


eliben:
Thanks for the attempt. I'm not 100% sure it fits the general flow of gobyexample, being so synthetic. Have to think more about this

PR #487, opened 2023-09-14
Added http-server /headers example

c0dehu1k:


eliben (commented):
Thanks. I'm not sure this is necessary - it doesn't really teach anything new about Go or the server, IMHO.

c0dehu1k:
Article talks about header and echoing various header options, so I thought will be useful to at least mention it as another curl command.

PR #489, opened 2023-09-16
Fix syntax example in the Variable page

siashish:
In the variable page, fix the syntax according to the example


eliben (commented):
The original text is the intended one. The comment explains what the code is equivalent to

PR #490, opened 2023-09-28
Update Atomic Counters example to use atomic.Uint64

erazemk:
This commit updates the Atomic Counters example to follow atomic's recommendation of using atomic.Uint64 instead of uint64 (same for other types) and then calling methods on it, instead of calling atomic.AddUint64().

It also updates the comments and empty lines a bit to look better on the website.


examples/atomic-counters/atomic-counters.go
```diff
@@ -30,12 +30,13 @@ func main() {
...
 		go func() {
 			for c := 0; c < 1000; c++ {
-				// To atomically increment the counter we
-				// use `AddUint64`, giving it the memory
-				// address of our `ops` counter with the
-				// `&` syntax.
-				atomic.AddUint64(&ops, 1)
+
+				// To atomically increment the counter we use `Add`,
+				// giving it the memory address of our `ops` counter
```
eliben, line 35: This comment is out of date now that the `Add` method is used

examples/atomic-counters/atomic-counters.go
```diff
@@ -48,5 +49,5 @@ func main() {
 	// atomics safely while they are being updated is
 	// also possible, using functions like
 	// `atomic.LoadUint64`.
-	fmt.Println("ops:", ops)
+	fmt.Println("ops:", ops.Load())
```
eliben, line 52: Here too, the comment it outdated

eliben (changes requested):
Thanks! It's about time we updated these, 1.19 was release a while ago

erazemk:
Ah sorry, only changed the function name, but forgot to change the rest of it. I've updated both comments to be more reasonable.

eliben (approved):
Thanks!

I may tweak some language later, but approving this PR for now

Issue #497, opened 2023-11-28
Add example: Operaions on closed channels


zhuboris:
I think that the [Closing Channels](https://gobyexample.com/closing-channels) page should include an explanation of what happens when you read from and write to a closed channel. I would like to add a short example at the end of the section.

eliben:
No objections in principle, but note that the example following that one is about ranging over a channel until it's closed

PR #498, opened 2023-11-28
Added example of reading from closed channel

zhuboris:
#497


zhuboris:
I've decided to not include example of writing to a closed channed because `panic` was never explained in previous examples.

examples/closing-channels/closing-channels.go
```diff
@@ -47,4 +47,10 @@ func main() {
 	// [synchronization](channel-synchronization) approach
 	// we saw earlier.
 	<-done
+
+	// It is possible to read from an empty closed channel,
+	// but instead of waiting for a message, we will
+	// immediately receive a zero value of the channel's type.
+	j := <-jobs
```
eliben, line 54: I think it'd be useful to demonstrate the 2-return form of `<-` here, because it's aimed exactly at this use case

examples/closing-channels/closing-channels.sh
```diff
@@ -7,6 +7,11 @@ sent job 3
 received job 3
 sent all jobs
 received all jobs
+no jobs to receive 0
 
-# The idea of closed channels leads naturally to our next
-# example: `range` over channels.
+# Be sure that channel is not closed when you read
```
eliben, line 12: Let's leave this out, keeping the original text.
gobyexample tries not to be too tutorial-y describing all the possible edge cases

zhuboris:
Сommitted the edits with an example of both returned values and revert the changes to the last paragraph

eliben (approved):

PR #499, opened 2023-12-07
fix: problem in burst limit

Shurtu-gal:
Suppose there is a sleep for some time then time.The tick channel keeps getting filled leading to the burst count > 3.

eg:- 
```go
  go func() {
    for t := range time.Tick(200 * time.Millisecond) {
      burstyLimiter <- t
    }
  }()

  time.Sleep(1000 * time.Millisecond)
```
image

Solution:- 
Drain the range by adding a default.
eg:- 
```go
  go func() {
    for t := range time.Tick(200 * time.Millisecond) {
      select {
        case burstyLimiter <- t:
        default:
          continue
      }
    }
  }()
  time.Sleep(1000 * time.Millisecond)
```
image


Shurtu-gal:
Yeah, like I had been using this in a sample app where I was rate-limiting the api-calls. However the problem I faced was, after some delay lets say after 500 seconds, the `time.Tick` channel has 500 entries which can lead to serve of 500 requests at once if they do come. 

Hence the need of drain.

eliben:
Thanks for the clarification.

I think this example is fine as it is. It provides the basic tools and concepts, without trying to solve every variation of the problem. The idea is expose Go's flexible building blocks for building solutions suitable for your specific use case.

Shurtu-gal:
Okay

PR #500, opened 2023-12-09
Update generics.sh with correct response

akhilpenugonda:
Update generics.sh with correct response
The response has to be [1 2 4].
image

### DIFF
--- a/examples/generics/generics.sh
+++ b/examples/generics/generics.sh
@@ -1,3 +1,7 @@
 $ go run generics.go
-keys: [1 2 4]
+keys: [4 1 2]
 list: [10 13 23]
+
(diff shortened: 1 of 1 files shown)


eliben (commented):
The order of iteration over map keys is not defined in Go, so different invocations get different orders

Would you like to update you PR to instead add a comment here in the `.sh` file saying something to that effect?

akhilpenugonda:
Thanks for letting me know, it helped. 
I added a comment instead of code change, please verify.

eliben (approved):

eliben:
Thanks, I will fix it up and update the public HTML

PR #504, opened 2024-01-18
Docs: Added Prev example for all the HTML files

1001pranav:
If we want to check for previous example then we had to click on home or back button for some might not work properly,

To Solve the above problem I have added prev anchor tag


eliben:
Please read https://github.com/mmcgrana/gobyexample/blob/master/CONTRIBUTING.md

Also, left arrow on the keyboard takes you to the previous example.

PR #508, opened 2024-02-09
Update waitgroups example to use go 1.22

pzerone:
goroutine closures inside for loops does not require variable reassignment since the 1.22 for loop changes


eliben (approved):
Nice find, thanks!

PR #510, opened 2024-02-22
Fix verbiage for Strings and Runes (noissue)

aannirajpatel:
Found a minor inaccuracy in the write-up for the "Strings and Runes" section: a UTF-8 "code point" can span multiple bytes for a single Thai character; a single Thai character won't span multiple UTF-8 code points because a "UTF-8 code point" is defined as a specific point on the UTF-8 character map, pertaining to a singular, specific character.


eliben (approved):
Thanks

PR #511, opened 2024-02-29
use blank import for importing the embed package

Yakiyo:
not using `_` fails compilation with an unused import error


Yakiyo:
sorry my bad, i think i messed something up. nevermind

Issue #491, opened 2023-09-29
Add Dark Mode 


soyreda:
is there any possibility for adding an option for switching to dark mode while using the website ?

Issue #476, opened 2023-05-20
Proposal: Adding dark theme


agilov:
Good day everyone! Often, I find myself reading gobyexample late into the night, sometimes even before going to bed. Given this, I'd like to propose the addition of a dark theme for the site, complete with automatic preference scheme detection, similar to what's in place on go.dev:

<img width="680" alt="go.dev dark mode switch" src="https://github.com/mmcgrana/gobyexample/assets/825624/6075f225-d75f-4b8e-b719-f389e445a4cf">

The actual color scheme can be modeled after the ones used on go.dev or github.com:

<img width="1205" alt="go.dev dark colors" src="https://github.com/mmcgrana/gobyexample/assets/825624/eba4fb9b-56c6-4940-9326-29b5a0076f97">

For the code colors, we can use those from the [Chroma Style Gallery](https://xyproto.github.io/splash/docs/). I recommend the "[github-dark](https://xyproto.github.io/splash/docs/all.html#github-dark)" style:

<img width="646" alt="code theme github-dark" src="https://github.com/mmcgrana/gobyexample/assets/825624/b239b481-bf85-4f54-8871-7ad8f73935cd">

If the maintainers are amenable to this idea, I am more than willing to implement it. It should be a straightforward task.

eliben:
Duplicate of #460

agilov:
@eliben, thank you for your response. I understand the concern regarding additional complexity in the frontend code. Having reviewed issue #460, I see how maintaining simplicity has been a priority for gobyexample.

However, I would like to highlight that the proposed feature would only require an additional 61 lines of CSS code, 4 lines of JavaScript, and 1 line of HTML. This is a relatively small addition considering the improved user experience for those who prefer a dark mode, particularly when reading at night.

Moreover, it's worth noting that many official Go resources have already implemented and maintained a dark mode. Having a similar feature on gobyexample could help maintain consistency across Go-related resources.

reignmaker:
The lack of the dark theme is the only thing that stops me from reading and recommending gobyexample, sadly.

PR #515, opened 2024-03-14
added: a dark mode in css for us night dwellers

ChrisMcKenzie:
This site has been a part of my reference site list for a long time but for me a dark theme user it gets a bit fatiguing so it would be nice to have a simple dark mode. hope it isn't offensive just thought it would be a nice addition.

the colors are based on tokyo_night

image

### DIFF
--- a/public/site.css
+++ b/public/site.css
@@ -42,6 +42,7 @@ table {
 	border-spacing: 0;
 }
 
+
 /* Layout and typography */
 body {
     font-family: 'Georgia', serif;
@@ -184,7 +185,7 @@ body .no { color: #880000 }                     /* Name.Constant */
 body .nd { color: #AA22FF }                     /* Name.Decorator */
 body .ni { color: #999999; font-weight: bold }  /* Name.Entity */
 body .ne { color: #D2413A; font-weight: bold }  /* Name.Exception */
-body .nf {  }                     /* Name.Function */
+body .nf {  }                                   /* Name.Function */
 body .nl { color: #A0A000 }                     /* Name.Label */
 body .nn { color: #0000FF; font-weight: bold }  /* Name.Namespace */
 body .nt { color: #954121; font-weight: bold }  /* Name.Tag */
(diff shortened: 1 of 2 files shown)


mmcgrana:
In the past I've resisted adding dark mode support because I wasn't a fan of any complexity, but I'm coming around to it, if we do it only in CSS as in this PR.

I think we should iterate on the exact styling:
* The white text on a dark background seems a little to thick to me, especially in the footer where we previously used grey-on-white.
* I'd want to look for a theme that was more black and less purple-oriented. Of the styles [here](https://pygments.org/styles/), `github-dark` and `one-dark` are the closest to me. I also think I had to tweak various values from the base theme when I did the light styling, but honestly it was so long ago I'm not sure.

If you're up to iterate on the styling per above please give it a shot. I also welcome any other observers to jump in with their own style suggestions.

I can't finalize the style myself right away but I can plan to get to it before too long, in the next week or so. 

The CSS file also needs some minor style and consistency tweaks but I'm happy to do that at the end.

Old issues for reference:
* #297 (in retrospect this was the right direction!)
* #460 (don't want to add a toggle button)
* #476 (recent proposal / request, mentions `github-dark`, cc @agilov)
* #491 (user asking about, issue closed out)

eliben:
Just a quick note that we use the Go package https://github.com/alecthomas/chroma for the syntax highlighting now. 

It seems to have a style gallery here: https://xyproto.github.io/splash/docs/longer/all.html with at least the `github-dark` style that @mmcgrana mentioned present

Of course we can customize it at will, I just wanted to make sure we don't get confused by pygments names in case things have deviated

ChrisMcKenzie:
great feedback I can work on this change tonight.

ChrisMcKenzie:
ok here is a screen of what I just added, sans purple and tweaks to the 

image

mmcgrana:
Thank you @ChrisMcKenzie!

I've built on your PR by further refactoring the code to make it easier for us to reason about light vs dark mode here and in the future. I've also done some further style experiments (see below).

I'd like to let this bake in PR for a bit to see if others have comments on the styling or anything else. I may also want to try more styling variations but this is all I can do today. IMO it's looking promising though!

Screenshot 2024-03-17 at 13 07 29

mon-jai:
@ChrisMcKenzie's second version looks better to me.

mmcgrana:
After thinking about it some more, I'm going to go ahead merge the iterated design I had with a few more tweaks.

We probably just all have our slightly different preferences, but fwiw what I'm trying to go for here a) a straight-greyscale background (vs blue or purple tinted) and subdued foreground colors (vs. too bright or neon), as we have in the light style. (Note that the screenshot may look unusually weird to you on GitHub if you have that on dark mode, because of the blue/purple tint on GitHub.)

Anyways part of this change makes it a lot easier for us to change syntax highlighting styles in the future, which I'll be open to doing (for both light and dark mode). Hopefully folks with dark mode enabled on their systems at least find this better than bright white 😎

PR #518, opened 2024-03-23
replace the clipboard/play icons

muhammadmp97:


eliben:
Thanks for the suggestion, but I think we're OK with the current icons

PR #519, opened 2024-03-27
Add the shell prompt/command to custom-errors.sh

johnd:
The shell entry for custom-errors is missing the shell prompt line. Small change, but nice for consistency!

### DIFF
--- a/examples/custom-errors/custom-errors.sh
+++ b/examples/custom-errors/custom-errors.sh
@@ -1,2 +1,3 @@
+$ go run custom-errors.sh
 42
 can't work with it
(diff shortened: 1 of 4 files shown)


johnd:
I had completely missed it there despite working through the examples back-to-back! I've added it in there and double checked all the other examples and they all seem to be fine.

eliben (approved):
Thanks!

EmNudge:
In both these examples, shouldn't the prompt be `custom-errors.go` and not `custom-errors.sh`?

eliben:
> In both these examples, shouldn't the prompt be `custom-errors.go` and not `custom-errors.sh`?

Good catch! Would you like to send a PR to fix this?

PR #521, opened 2024-04-10
Feature/nav

apoddaturi:
Adds Sentence to assist Users about navigation between examples using Right and left arrows.


eliben:
Considered this, but I think we're OK without for now

PR #524, opened 2024-04-26
Use seek enums and show other seek methods

kauppie:
- Converted `io.Seek(0, 0)` calls to `io.Seek(0, io.SeekStart)` to reduce the use of magic values.
- Added example of using `io.Seek(x, io.SeekCurrent)` and `io.Seek(x, io.SeekEnd)`.


eliben (approved):
LGTM, thank you

PR #526, opened 2024-05-01
Change ending extension from .sh to .go for errors lessons

EmNudge:
From https://github.com/mmcgrana/gobyexample/pull/519#issuecomment-2085150541


eliben (approved):
Thanks!

PR #527, opened 2024-05-11
[bugfix] Stop breaking browser navigation shortcuts

HSPDev:
On Linux Chrome I use ALT + Arrows to go back in the browser, but instead it paginates between examples.

Fixed it by checking for special key modifiers and do nothing (to the let the browser handle it). 
I think the user always intends a system shortcut if they use a modifier key.


eliben:
Thanks!

PR #530, opened 2024-05-14
Fix typo in the enums example

agorkem:
Should be `ServerState` instead of `StateType` on [Enums](https://gobyexample.com/enums)


eliben (approved):
Thanks!

PR #536, opened 2024-06-01
Shorten interface explanation

kkeian:
The explanation for embedding a struct which implements an interface is confusing.
The code for it restricts the `container` variable to one method: `describe`.

The example code in this patch uses the `describer` interface to specify a contract between a function and it's caller.
This explicitly gives a practical reason for using struct embedding or adding a method to an already embedded struct.


eliben (commented):
Thanks for the contribution.

I've read this through a couple of times and I think I like the original version more

PR #539, opened 2024-06-18
Update for.go - Untyped Int Constant

KeithGrennier:
Code as written gives an error of: "cannot range over 3 (untyped int constant)" and cannot range over 6 (untyped int constant)

Updated change seems to work, but I'm still learning go 😄


eliben (changes requested):
See https://github.com/mmcgrana/gobyexample?tab=readme-ov-file#what-version-of-go-is-required-to-run-these-examples

and try it with the newest release of Go

PR #540, opened 2024-07-03
Remove unreachable code from enums example

BigJayToDaIzo:
After confirming I did not misunderstand how switch default and panic works, I realized there was some unreachable code in the example that may confuse new learners.  Go LSP confirms.

image


eliben (approved):
Thanks, this makes sense!

BigJayToDaIzo:
> Thanks, this makes sense!

My pleasure!  Thank YOU for the lovely refresher course after not having looked at any Go for a month! :grin:

BigJayToDaIzo:
woops, fat fingered the mouse on that early close before merge.  Apologies!

eliben:
> woops, fat fingered the mouse on that early close before merge. Apologies!

No worries, it's merged now.

Note that it may take a few days to show up on the published website, since the upload is done manually.

PR #543, opened 2024-07-20
Added jagged array

KhanbalaRashidov:


examples/arrays/arrays.go
```diff
@@ -59,4 +59,36 @@ func main() {
 		{1, 2, 3},
 	}
 	fmt.Println("2d: ", twoD)
+
+	// Create a jagged array (slice of slices)
+	jaggedArray := [][]int{
```
eliben, line 64: Sorry, I'm not sure what this is adding.

It doesn't belong in arrays, but maybe in slices. But in slices we already have a multi-dimensional slice example.

KhanbalaRashidov: There is yes in slice part, but there was no example with array

PR #545, opened 2024-07-30
Added multiple consts

KhanbalaRashidov:


eliben (commented):
Thanks for the contribution. This is something I want to think about. I'm not opposed to having this sample, but I'd like it to be a bit simpler/shorter and fix the spelling of the comments.

KhanbalaRashidov:
> Thanks for the contribution. This is something I want to think about. I'm not opposed to having this sample, but I'd like it to be a bit simpler/shorter and fix the spelling of the comments.

Okay. Thank you

eliben:
Having thought some more about it, I don't think this is necessary for now

PR #549, opened 2024-08-11
Renaming instances of `http-clients`, `http-servers`, and `context-in-http-servers` to be consistent with examples.

bustosalex1:
- Renaming some remaining instances of `http-clients` and `http-servers` to be `http-client` or `http-server` to be consistent with the examples.
- Renaming `context-in-http-servers.go` in `context.sh` to be consistent with `context.go`


examples/context/context.sh
```diff
@@ -1,5 +1,5 @@
 # Run the server in the background.
-$ go run context-in-http-servers.go &
+$ go run context.go &
```
eliben, line 2: This change is worthwhile, but the http-server(s) http-client(s) - I prefer to keep as is

eliben (changes requested):

examples/context/context.sh
bustosalex1: gotcha. Just reverted those changes.

eliben (approved):

PR #551, opened 2024-08-21
Update arrays.go updating syntax for declaring and initializing an array in one line

CooperSawyer1:
Small nit but in line 36 and 41 the colon is missing for  declaring and initializing an array in one line. Just adding that in.

### DIFF
--- a/examples/arrays/arrays.go
+++ b/examples/arrays/arrays.go
@@ -33,12 +33,12 @@ func main() {
 
 	// You can also have the compiler count the number of
 	// elements for you with `...`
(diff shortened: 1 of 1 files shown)


eliben (commented):
Did you try running the code you're suggesting?

CooperSawyer1:
> Did you try running the code you're suggesting?

I did. It is also shown in the example above for 
`b := [5]int{1, 2, 3, 4, 5}`

CooperSawyer1:
I see my mistake now. I was running the commands separately without the array declared like you did above. Misunderstanding on my part.

PR #554, opened 2024-09-22
Correct inaccuracy in recursion description.

wwwstephen:
An example of recursion in an anonymous function is given, but in the comments it describes the anonymous function as a closure. I found this confusing on reading this. I have made the necessary changes.


eliben (approved):
Makes sense, thank you

PR #555, opened 2024-09-23
Fixed missing command in the range-over-iterators file

wwwstephen:
The `range-over-iterators` file does not include the command `go run range-over-iterators.go`, which is inconsistent with the rest of the project. This omission may confuse users, as they won’t know what to name their file or which command to run locally. I have fixed this.

### DIFF
--- a/examples/range-over-iterators/range-over-iterators.sh
+++ b/examples/range-over-iterators/range-over-iterators.sh
@@ -1,3 +1,4 @@
+$ go run range-over-iterators.go
 10
 13
 23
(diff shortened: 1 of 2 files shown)


eliben (approved):

eliben:
This looks fine, but can you please squish to a single commit?
This seems to mix up commits from a different PR and I want to tread carefully

wwwstephen:
OK, do you know what's the best way to do that? Will I just create a new PR?

wwwstephen:
OK, I've tried a few things and think the commit is now in correct order :-) Every day is a school day.

eliben (approved):

PR #557, opened 2024-10-26
clarify impicit implementation of interface 

chetan6780:
Go type implements interfaces automatically and there is no implements keyword, this is very useful for people coming from other programming languages.

Preview:
image


eliben (commented):
Thanks for the suggestion, but I don't think this clarifies much for people who don't come from a language that has an "implements" keyword. For example, I never saw such a keyword before finding Go myself.

At the end of the day, gobyexample is not a first tutorial - it's meant to be consulted after one has seen a more introductory tutorial like the tour of Go

PR #566, opened 2024-12-29
Added json.NewDecoder example

bhaskar253:
adding example for json.NewDecoder


eliben (approved):
Thanks, I think this is useful to have!

I'm going to merge this and will then adjust the comment text a bit

Issue #523, opened 2024-04-16
Add example of type assertions


adriancuadrado:
Resources:
* https://go.dev/ref/spec#Type_assertions
* https://go.dev/tour/methods/15

kevin-kho:
> This could be useful if a good example can be worked out. There's some related code in the `switch` and `json` examples already, but maybe this would go to `interfaces`

I recommend this to go into `interfaces`. The perimeter of a circle is also called a circumference so a `circum()` method could be defined on the `circle` struct that does the same as `perim()`. Do type assertion against a `shape` to check if it's `circle` or `rect` before calling either `perim()` or `circum()`

Example below

```go
// alias for perim()
func(c circle) circum() float64 {
    return c.perim()
}


shapeSlice := []geometry{rect{width: 5, length: 10}, circle{radius: 5}}

for _, shape := range shapeSlice {

    fmt.Println(shape.area())

    if r, ok := shape.(rect); ok {
       fmt.Println(r.perim())
    }

    if c, ok := shape.(circle); ok {
       fmt.Println(r.circum())
    }

}

```

akalanitski:
Could Go by Example have a separate example for Type Assertions, instead of including it as part of other examples?
I’ve prepared a PR https://github.com/mmcgrana/gobyexample/pull/586 for discussion

eliben:
This issue should have been closed when #567 landed. I think this is sufficiently demonstrated now, and a separate example isn't required

PR #567, opened 2025-01-03
added type assertion example

kevin-kho:
addresses issue #523


eliben (commented):
Thanks for the PR.

I would simplify it more. We don't need the `circum` method.

I'd just create a function called something like `detectCircle` that takes a `geometry` and prints out its radius it's a cirlce. The comment should clarify that this is done if we want to explicitly check the runtime type of the value; the comment should also point to the switch example for an alternative approach.

kevin-kho:
> Thanks for the PR.
> 
> I would simplify it more. We don't need the `circum` method.
> 
> I'd just create a function called something like `detectCircle` that takes a `geometry` and prints out its radius it's a cirlce. The comment should clarify that this is done if we want to explicitly check the runtime type of the value; the comment should also point to the switch example for an alternative approach.

addressed in commit [6d59536](https://github.com/mmcgrana/gobyexample/pull/567/commits/6d595368ff21a8cd569c98d519c4cb3e4a4969ea)

eliben (approved):
Thank you - this is close enough to what I want that I'll go ahead and merge it. I'll tweak it a bit before republishing

Issue #568, opened 2025-01-06
Add rel="next" to next example link


yodigi7:
Add rel="next" to next example link. This will help with tools such as vimium to navigate to the next page.

eliben:
I'm not familiar with `rel="next"`. Some online searching points out to it being mainly for SEO and it should be placed in the document `<head>`, not on specific `<a ...>` elements. Can you point to some resources that explain the usage you're describing here?

yodigi7:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#next

yodigi7:
I did build the site locally and did confirm that my vimium extension did recognize it as a "next" link.

eliben:
> https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#next

Thanks!

eliben:
Fixed by #569

PR #569, opened 2025-01-06
Update example.tmpl to use rel="next" on next link

yodigi7:
Fix for issue #568.

### DIFF
--- a/public/arrays
+++ b/public/arrays
@@ -217,7 +217,7 @@ when printed with <code>fmt.Println</code>.</p>
       
       
       <p class="next">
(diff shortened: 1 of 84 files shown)


yodigi7:
Note: the main change is in the example.tmpl file.

eliben (approved):
Can you get the test to pass?

Run `tools/build` with the env `TESTING=1` set locally

yodigi7:
Yeah, I can work on that

yodigi7:
@eliben ok, think I got it passing, not seeing the CI run anymore but I think it should be good. Had a weird build I pushed up initially but should be working now.

PR #570, opened 2025-01-28
Added simpler example and explanation for use of defer in go

gentcod:
This change adds more description and explanation to the flow of execution and use of `defer` in Go.
It is a simple representation of how `defer` statement alters execution, using just logged strings.

### DIFF
--- a/examples/defer/defer.go
+++ b/examples/defer/defer.go
@@ -2,6 +2,8 @@
 // performed later in a program's execution, usually for
 // purposes of cleanup. `defer` is often used where e.g.
 // `ensure` and `finally` would be used in other languages.
+// `defer` enables execution to be run in
+// Last-In-First-Out (LIFO) order.
 
 package main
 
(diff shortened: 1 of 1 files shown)


examples/defer/defer.go
```diff
@@ -15,11 +17,20 @@ import (
 // do that with `defer`.
 func main() {
 
+	// This enclosing anonymous function will be executed at the end
```
eliben, line 20: I'm not sure what you mean by "simpler", and what realistic use case does this demonstrate?

gentcod: While it doesn’t represent a real-world use case, it provides a clear and minimal demonstration of how deferred calls are executed in LIFO order, making it easier to understand before applying defer in practical scenarios like resource cleanup.

eliben:
Thanks, I understand the reasoning, but I think the current example is better aligned with the overall goal of gobyexample

PR #571, opened 2025-02-04
clarify a comment w.r.t. empty slice

hudem1:
Hi !

I'm reading through gobyexample these days as I'm learning Go, and I came across 2 things:
- i think there were forgotten `go` keywords for calling routines.
- i wanted to suggest a slight text adjustment, which I had a bit of hard understanding at first. I think when creating a slice like `s = make([]string, 3)`, the slice is not empty, it contains 3 elements, the only thing is those elements are zero-valued but still the slice is not empty. I just thought it was a bit confusing when I read that, I had to read several times and then test it to really make sure I understood. So, here is a very little adjustment.

Let me know what you think! :)


examples/channel-directions/channel-directions.go
```diff
@@ -24,7 +24,7 @@ func pong(pings <-chan string, pongs chan<- string) {
 func main() {
 	pings := make(chan string, 1)
 	pongs := make(chan string, 1)
-	ping(pings, "passed message")
```
eliben, line 27: Because the channels are buffered, the `go` statements aren't necessary here. They won't hurt, but one may wonder if they're needed. 

I suppose the channels could be made unbuffered and `go` keywords added, this would have a similar effect

examples/slices/slices.go
```diff
@@ -17,9 +17,9 @@ func main() {
 	var s []string
 	fmt.Println("uninit:", s, s == nil, len(s) == 0)
 
-	// To create an empty slice with non-zero length, use
+	// To create a slice with non-zero length, use
```
eliben, line 20: I think this is ok, changing "an empty" to "a", but the `ie ""` below is an unnecessary addition

examples/channel-directions/channel-directions.go
hudem1: Yes okay, it's just that I thought without `go` the code just becomes synchronous, so we're guaranteed that `ping` will happen before `pong` before the main exiting. But with adding `go`, it's the channels directions, especially receiving from channels that enforces the order of execution (`ping` then `pong` then the main routine exiting). So, I thought adding `go` really made clear that thanks to blocking when trying to receive from channels, we get the correct order of execution.

But I guess here the example was mostly about the fact that we can pass specific channel directions as function parameters. So, the underlying example matters a bit less.

Let me know what you'd prefer, I can remove the `go` if you think it's better this way, no worries :)

examples/slices/slices.go
hudem1: Yes okayy sure!

examples/channel-directions/channel-directions.go
eliben: Yeah let's just remove the `go`s

eliben:
Thanks for the adjustment.

Would you mind running `tools/build` in the root directory and adding the generated file to the PR? This will enable me to just merge it directly without fixups

(https://github.com/mmcgrana/gobyexample/blob/master/CONTRIBUTING.md)

hudem1:
> Thanks for the adjustment.
> 
> Would you mind running `tools/build` in the root directory and adding the generated file to the PR? This will enable me to just merge it directly without fixups
> 
> (https://github.com/mmcgrana/gobyexample/blob/master/CONTRIBUTING.md)

Oops, sorry forgot to do that!

PR #592, opened 2025-05-06
fix: order of generics

tarasowski:
Fix type parameter order in SlicesIndex for clarity

Reordered type parameters so that E is defined before S (~[]E), making the dependency clear. The previous order (S before E) was confusing for beginners, as S depends on E being defined. This improves readability and aligns with best practices in Go generics.

### DIFF
--- a/examples/generics/generics.go
+++ b/examples/generics/generics.go
@@ -14,7 +14,7 @@ import "fmt"
 // of this type signature, see [this blog post](https://go.dev/blog/deconstructing-type-parameters).
 // Note that this function exists in the standard library
 // as [slices.Index](https://pkg.go.dev/slices#Index).
-func SlicesIndex[S ~[]E, E comparable](s S, v E) int {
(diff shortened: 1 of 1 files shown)


eliben:
What makes you say this aligns with best practices? Can you point to specific examples?

Note that the official `slices` package is written this way also, e.g. see https://pkg.go.dev/slices#Index
Also the official Go blog posts (e.g. https://go.dev/blog/generic-slice-functions)

tarasowski:
As I mentioned in my first issue, it can be a bit confusing for beginners. But since this is the idiomatic way in Go, I'm happy to close the issue.
I was just exploring whether both approaches would work—and they do, at least in the examples I tested (e.g., ...).

```go
func Index[E comparable, S ~[]E](s S, v E) int {
    for i := range s {
        if v == s[i] {
            return i
        }
    }
    return -1
}
```

PR #593, opened 2025-05-06
Small miss consistency in file names

akalanitski:
All files in repo names according to the simple convention. 
I have found 3 miss consistency in file name, please check PR


eliben (changes requested):
The testing and benchmarking change will not work - this needs to demonstrate the `_test.go` convention for the tests to work.

`http-server.hash` has already been fixed in a separate commit. I'll fix `http-client.hash` similarly.

PR #601, opened 2025-05-10
Fix Markdown syntax for links ended by ")"

akalanitski:
Some examples used HTML links instead of Markdown because of "(" and ")" are conflicting with MD-syntax.
I escape the parentheses  in URLs and rewrite links in MD style.


eliben (commented):
Thanks, but I don't think this makes things better. Now copy pasting the link from the code doesn't work, and any new link has to be adjusted this way. There's no problem with having HTML-style links in some cases

PR #615, opened 2025-07-18
Fix typo in function name

DeinAlptraum:


eliben (commented):
Can you clarify why this is a typo?

DeinAlptraum:
The code example right after that contains the function call `http.HandleFunc`, and I've confirmed that that one works.

PR #621, opened 2025-08-28
Ignore arrow key press if the Command key is held on MacOS

JacobThwaites:
This PR includes the `metaKey` (the "Command" key on Mac) in the modifier keys check when handling the keydown event. This fixes a bug for navigating when using MacOS. Specifically, the MacOS shortcut for navigating history in the browser is `cmd+left/right`, however currently this leads to unexpected behaviour where it sometimes navigates according to the browser history, and other times according the site's internal navigation logic.

With these changes, the command key press on MacOS is now handled as expected.

### DIFF
--- a/public/arrays
+++ b/public/arrays
@@ -7,7 +7,7 @@
   </head>
   <script>
       window.onkeydown = (e) => {
-          if (e.ctrlKey || e.altKey || e.shiftKey) {
+          if (e.ctrlKey || e.altKey || e.shiftKey || e.metaKey) {
               return;
           }
           
--- a/public/atomic-counters
+++ b/public/atomic-counters
@@ -7,7 +7,7 @@
   </head>
   <script>
       window.onkeydown = (e) => {
-          if (e.ctrlKey || e.altKey || e.shiftKey) {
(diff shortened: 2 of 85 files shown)


eliben:
I cannot reproduce this. On mac OS, cmd-left correctly uses browser history for me

JacobThwaites:
It doesn't happen for me all the time. It seems to happen consistently if you first navigate through the pages using just the arrow keys, then try using `cmd-left`. For example:

1. Open a new browser tab
2. Go to https://gobyexample.com/constants
3. Navigate to the next page using the right arrow key (should take you to the "for" page)
4. Now keep pressing `cmd-left` to navigate back through browser history.

For me, this cycles back through all the pages (even ones I haven't viewed), ending in a cycle between the "hello-world" and "values" page.

eliben (approved):

eliben (approved):

eliben:
Thanks for the reproducer. What I think is happening is that the browser runs out of actual history, so it doesn't intercept the cmd+left any more and it goes to the window. So this is quite a corner case!

That said, the change looks safe enough to me because of the existing checks for other key modifiers. So I'll merge it.

Thanks for the contribution

PR #623, opened 2025-09-07
(minor): update mutexes example to use wg.Go instead of manual creation

Dawo9889:
This PR contains an update proposal to use wg.Go function in the mutexes section, as showed in an example of WaitGroups section. This function was introduces in go 1.25 https://pkg.go.dev/sync#WaitGroup.Go


Dawo9889:
Oh, Sorry for forgetting about that.
Should be good now

PR #625, opened 2025-09-27
Improve Spawning Processes, using errors.As

obaibula:
In Go 1.13, the new function errors.As was added, which examines the error tree looking for an error that can be assigned to the target. Using As is preferable to the approach shown in Go by Example, because As will succeed even if the error is wrapped.

Note: In this example the error is not wrapped, but since Go by Example is a learning resource, the recommended way of examining errors should be demonstrated.


examples/spawning-processes/spawning-processes.go
```diff
@@ -30,16 +31,22 @@ func main() {
...
-	// if the command ran but exited with a non-zero return
-	// code.
+	// command (e.g. executable file not found in $PATH),
+	// and `*exec.ExitError` if the command ran but
+	// exited with a non-zero return code.
 	_, err = exec.Command("date", "-x").Output()
 	if err != nil {
-		switch e := err.(type) {
-		case *exec.Error:
+		var (
```
eliben, line 39: I'd just do:

```
var execErr *exec.Error
var exitErr *exec.ExitError

```

IMHO saving one `var` keyword doesn't justify the two extra lines

Issue #562, opened 2024-11-28
Error executing code due to file path on windows


Yudaev:
https://github.com/mmcgrana/gobyexample/blob/bc680ed6b2a7b2fab8a511242cccfdbc5a78b100/examples/defer/defer.go#L23

This code does not work in Windows.
Corrected universal code:
`
f := createFile(filepath.Join(os.TempDir(), "defer.txt"))
`

eliben:
Can you open a PR?

PR #629, opened 2025-11-02
fix issues : Error executing code due to file path on windows #562

marktech0813:
f := createFile("/tmp/defer.txt")  : This code does not work in Windows.
Corrected universal code:
f := createFile(filepath.Join(os.TempDir(), "defer.txt"))


marktech0813:
If you’re hitting a compile error in a branch or a different example, add the missing import:
  import "path/filepath"

PR #633, opened 2025-11-13
feat(example.tmpl): Add previous navigation to example pages

JiruGutema:
**Add previous navigation to example pages**

Add an explicit "Previous example" link/button to the example template so
each example page shows a backwards navigation link in addition to the
existing keyboard navigation. The link uses rel="prev" for semantics.

Files changed:
- templates/example.tmpl
- templates/site.css

Note: I was using this website and needed a button to go back, so I added it now. 
image

### DIFF
--- a/templates/example.tmpl
+++ b/templates/example.tmpl
@@ -25,6 +25,11 @@
   <body>
     <div class="example" id="{{.ID}}">
       <h2><a href="./">Go by Example</a>: {{.Name}}</h2>
+    {{if .PrevExample}}
+      <p class="prev">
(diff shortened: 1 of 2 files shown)


JiruGutema:
> Hello, I think a similar PR was created regarding the same: #139

I’ve been actively using the website myself, and I kept struggling to go back to the previous example without relying on keyboard navigation. It wasn’t obvious or convenient on the interface, so I ended up adding a visible “Previous example” button to improve usability.

I hope this clarifies the motivation behind the change. Let me know if you'd like adjustments or if it makes sense to merge with the earlier PR.

JiruGutema:
What is the logic behind it tho? why you guys are not allowing it?

PR #634, opened 2025-11-14
Fix UI code copying bug

stefali1-dev:
# Fix text selection bug by replacing table layout with CSS Grid

## Problem

When selecting code snippets with the mouse on example pages, unintended documentation text at the same height was also being selected.

### Root Cause

The issue seemed to be from the HTML table-based layout (`<table>`, `<tr>`, `<td>`) used to display code and documentation side-by-side.

## Solution

This PR replaces the table-based layout with **CSS Grid**, which provides proper column isolation for text selection. The key changes are:

### Template Changes (`templates/example.tmpl`)
Lines changed: 11 additions & 11 deletions

- Replace `<table>` with `<div class="seg-grid" role="presentation">`
- Replace `<tr>` and `<td>` elements with semantic `<div>` elements
- Add `style="grid-row: {{.Row}};"` to position code and docs segments correctly
- Render code segments first, then docs segments (both using their respective grid rows)

### CSS Changes (`templates/site.css`)
Lines changed: 16 additions & 24 deletions

- Define `.seg-grid` as a CSS Grid container with two explicit columns:
  - Column 1 (420px): Documentation
  - Column 2 (480px): Code
- Update all selectors from `td.docs`/`td.code` to `.seg-grid .docs`/`.seg-grid .code`
- Use `grid-column` property to position elements instead of relying on table cell structure
- Remove obsolete table-related CSS rules (`table td`, `vertical-align`, etc.)
- Clean up unused syntax highlighting selectors (`.nx`, `.nf`, `.o`, `.p`) that had no style declarations

### Code Changes (`tools/generate.go`)
Lines changed: 2 additions & 0 deletions
- Add `Row int` field to the `Seg` struct to track segment positioning
- Populate `Row` field during segment parsing (`seg.Row = i + 1`)

## Testing

- Built the site locally with `tools/build` & verified text selection behavior works correctly (code and docs are independently selectable)
- Confirmed visual layout is unchanged
- Tested in both light and dark modes

### DIFF
--- a/public/arrays
+++ b/public/arrays
@@ -26,181 +26,164 @@
     <div class="example" id="arrays">
       <h2><a href="./">Go by Example</a>: Arrays</h2>
       
-      <table>
+      <div class="seg-grid" role="presentation">
         
-        <tr>
-          <td class="docs">
-            <p>In Go, an <em>array</em> is a numbered sequence of elements of a
-specific length. In typical Go code, <a href="slices">slices</a> are
-much more common; arrays are useful in some special
-scenarios.</p>
-
-          </td>
-          <td class="code empty leading">
-            
+        <div class="code empty leading" style="grid-row: 1;">
           
-          </td>
-        </tr>
         
-        <tr>
-          <td class="docs">
-            
-          </td>
-          <td class="code leading">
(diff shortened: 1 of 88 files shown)


eliben:
Thanks, but we won't be making a massive change to the site's layout. I don't have the time and resources to test it works well on all the different devices and configurations people use to view gobyexample; the current layout has served it well enough for a long time, and the copy-paste issue is trivially mitigated by the existing buttons in the top-right corner of each sample.

PR #631, opened 2025-11-05
Added TCP and UDP examples on Socket Programming

0xprincedev:
## Description

Added TCP and UDCP examples on Socket Programming

## Testing

- [x] Code compiles without errors
- [x] Examples follow the project's format and style
- [x] Comments are comprehensive and educational
- [x] Shell scripts demonstrate expected output
- [x] Tested examples run successfully

## Checklist

- [x] Examples use only the standard library
- [x] Code follows Go conventions and project style
- [x] Comments explain concepts clearly for beginners
- [x] Examples are added to `examples.txt`
- [x] Shell scripts (`*.sh`) are included with expected output
- [x] No linter errors

Fixes #619

PR #636, opened 2025-11-18
Add Basic TCP Server Example (Fixes #619)

### LINKED ISSUE #619: Proposal: Adding example on Socket Programming
TCP and UDP protocols play crucial rules in most of  common soft wares , writing clean and perfect in socket programming is a quiet challenging task for beginners .
[sample](https://github.com/Omid-SH/Web_Final_Issues/blob/main/%23316/server.go)

bhushan-mdn:
Hello,

This Pull Request introduces a focused implementation for a basic TCP Server, directly addressing **Issue #619** and incorporating the feedback provided in the review of **PR #631**.

The goal is to provide the simplest, most educational example of network programming with the `net` package for beginners.

## Implementation Details

- **Simplicity and Focus**: The example consists of a single server implementation (`tcp-server.go`) that handles concurrent connections using goroutines.
- **Testing**: Aligned with project suggestions, the accompanying script (`tcp-server.sh`) uses the standard system utility netcat (`nc`) for testing the connection and response cycle.

## Request for Input on Scope

Before I submit further networking examples, I would appreciate guidance on the overall scope:

1. Is this sufficient? Should we merge this minimal TCP Server example now?
2. Should I expand the set? Would you prefer a consolidated approach in a single future PR that adds other primitives (e.g., TCP Client, UDP Server/Client)?

Thank you for your time and review!

### DIFF
--- a/examples/tcp-server/tcp-server.go
+++ b/examples/tcp-server/tcp-server.go
@@ -24,8 +24,7 @@ func main() {
 
 	// Loop indefinitely to accept new client connections.
 	for {
-		// `listener.Accept()` blocks the loop until a client
-		// attempts to connect, returning a net.Conn.
+		// Wait for a connection.
 		conn, err := listener.Accept()
 		if err != nil {
 			log.Println("Error accepting conn:", err)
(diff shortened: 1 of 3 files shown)


bhushan-mdn:
Hi @eliben ,

I've pushed the requested changes:
* Simplified the comments around `listener.Accept()` & `handleConnection` and the main loop to focus on *purpose* rather than *implementation* (removed goroutine mention).
* Removed unnecessary comment above log.

Please let me know if it looks good. Thanks again for the feedback!

PR #651, opened 2026-03-13
Change declaration of twoD to use short variable syntax

gksoriginals:
Fix undefined: twoD error


dnp314:
Hi @gksoriginals, the twoD variable has already been declared above: `var twoD [2][3]int`
so using `:=` would throw an error:
`./prog.go:57:7: no new variables on left side of :=`

PR #657, opened 2026-05-14
fix: check the error returned by filepath.WalkDir

malklera:


eliben (approved):

eliben:
Thanks for the contribution!

PR #658, opened 2026-05-15
fix(examples): migrate signals example from syscall to golang.org/x/sys/unix

### LINKED ISSUE #656: Update Go by Example: Signals
## Signals example uses deprecated `syscall` package

The [signals example](https://gobyexample.com/signals) currently uses `syscall` to reference os signals constants.

The `syscall` package is frozen and its own documentation recommends migrating away from it:

```
NOTE: Most of the functions, types, and constants defined in this package are also available in the [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) package. That package has more system call support than this one, and most new code should prefer that package where possible. See https://golang.org/s/go1.4-syscall for more information.
```

### Suggested change

Replace the `syscall` import with `golang.org/x/sys/unix`.

algojogacor:
## Summary

Closes #656 - migrates the signals example away from the deprecated syscall package.

## Problem

The signals example uses syscall to reference signal constants (syscall.SIGINT, syscall.SIGTERM). The syscall package is frozen and its documentation recommends migrating to golang.org/x/sys.

## Changes

- examples/signals/signals.go: Replace syscall import with golang.org/x/sys/unix. Use unix.SIGINT and unix.SIGTERM instead of their syscall counterparts.
- go.mod: Add golang.org/x/sys v0.38.0 as a direct dependency.

## Verification

The unix.SIGINT and unix.SIGTERM constants have identical integer values to their syscall counterparts, so the program behavior is unchanged.

### DIFF
--- a/examples/signals/signals.go
+++ b/examples/signals/signals.go
@@ -11,14 +11,15 @@ import (
 	"context"
 	"fmt"
 	"os/signal"
-	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 func main() {
 	// `signal.NotifyContext` returns a context that's canceled
 	// when one of the listed signals arrives.
 	ctx, stop := signal.NotifyContext(
-		context.Background(), syscall.SIGINT, syscall.SIGTERM)
+		context.Background(), unix.SIGINT, unix.SIGTERM)
 	defer stop()
 
 	// The program will wait here until one of the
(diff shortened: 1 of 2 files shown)


eliben (approved):

eliben:
Please follow https://github.com/mmcgrana/gobyexample/blob/master/CONTRIBUTING.md to re-generate the HTML in the `public` dir from example sources

PR #662, opened 2026-06-07
fix: check unmanaged errors

malklera:
Add checking to the errors returned by json.Unmarshal(), *Encoder.Encode(), *Decoder.Decode().

Leave the ignored(_) errors as is, I suppose they are being managed.


eliben (changes requested):
Not sure about this... there's something to be said about shorter, cleaner code for this example since by this time in gobyexample users should be familiar with the general concept of error handling.

I'd accept a PR that adds a bit of text in the comment on the first use of errors (from "Here’s the actual decoding, and a check for associated errors....") saying something about doing this for all json package calls

malklera:
Replaced the handling of the errors to ignore them.
Let me know if the phrasing in the added comment is ok.

eliben (approved):

PR #664, opened 2026-07-09
Small refactor and var rename

LaurTudo:
Hi!

While using tools/build to translate the site (PR coming in the next few weeks, hopefully), I noticed two small things in tools/measure.go, so I made the following changes:

- replaced the magic number 68 with a maxLineLength constant
- renamed commentPat to commentPattern for clarity


eliben (approved):

eliben:
Thanks!

PR #665, opened 2026-07-11
Use any in switch example

comick:
`any` is an alias for `interface {}` which could be simpler for an early example such as the switch one.


examples/switch/switch.go
```diff
@@ -44,10 +44,10 @@ func main() {
 	}
 
 	// A type `switch` compares types instead of values.  You
-	// can use this to discover the type of an interface
+	// can use this to discover the type of an any interface
```
eliben: Please leave the comment as is

comick: done

eliben (approved):

PR #670, opened 2026-07-19
add missing fmt.Println outputs in pointers example

### LINKED ISSUE #669: Missing output lines in Pointers example
The last two `fmt.Println` statements don't show their output
Expected 6 lines of output, but only 4 are displayed

A-Lokanush:
Fixes #669 

image


eliben (commented):
Thanks!

PR #672, opened 2026-07-29
Added EXAMPLE_FILE_SUFIX and .FileNameSufix

kberov:
I have a cheap shared hosting account served by Apache.
I want to upload my generated files from the public folder to an arbitrary folder under my domain name.

I do not want to add specific for this folder .htaccess file to serve the produced static files without file extensions in the arbitrary folder.
I want Apache or any other server to simply serve the *.html files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,14 @@ To build the site you'll need Go installed. Run:
 $ tools/build
 ```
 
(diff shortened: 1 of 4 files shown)


eliben (commented):
Sorry, I think you should maintain this in your own fork

kberov:
Ok, thanks.

PR #674, opened 2026-08-01
using golang.org/x/sys/unix over syscall package

### LINKED ISSUE #656: Update Go by Example: Signals

dnp314:
Fixes #656 and even `execing-processes` was using syscall
#658 PR already exists, but seems to be inactive


eliben:
Closing, please see the new comment in #656

PR #676, opened 2026-08-12
Update json example to encoding/json/v2

prattmic:
The core marshal/unmarshal APIs are the same, so most examples don't change at all.

MarshalWrite and UnmarshalRead provide an easier way to use io.Reader/Writers.

Go 1.27 is not released yet. This shouldn't be merged until some time after release.


examples/json/json.go
```diff
@@ -120,16 +120,15 @@ func main() {
...
 	// strings as intermediates between the data and
 	// JSON representation on standard out. We can also
-	// stream JSON encodings directly to `os.Writer`s like
+	// stream JSON encodings directly to `io.Writer`s like
 	// `os.Stdout` or even HTTP response bodies.
-	enc := json.NewEncoder(os.Stdout)
 	d := map[string]int{"apple": 5, "lettuce": 7}
-	_ = enc.Encode(d)
+	_ = json.MarshalWrite(os.Stdout, d)
+	fmt.Println()
```
prattmic: This is awkward, I'm not sure the best way to handle this.

encoding/json.Encoder writes a newline after the JSON output. encoding/json/v2.MarshalWrite does not, so without an extra newline here this example and the next are output on the same line.

encoding/json/jsontext.Encoder (to be used with json.MarshalEncoder) _does_ write a newline, but I don't think we want that here, since MarshalWrite is what most users will want most of the time.

prattmic: I suppose one alternative would be to MarshalWrite to a bytes.Buffer and then print the buffer afterwards.

prattmic: > I suppose one alternative would be to MarshalWrite to a bytes.Buffer and then print the buffer afterwards.

Done. I think this is pretty natural given that the next example also uses `strings.Reader`.

eliben:
LGTM. I plan to merge this after 1.27.1 is released to give folks some time to absorb 1.27

PR #677, opened 2026-08-20
Fix invalid new(42) syntax in pointers example and rebuild

lukettoOoO:
Fixes a syntax error where `new(42)` was used instead of allocating an int type (`new(int)`). 

Regenerated the site using `tools/build`.


eliben (commented):
See https://github.com/mmcgrana/gobyexample/commit/b868b950ed3245457134c50e528ea8a9d81ae80a

dnp314:
We can use the `new` function with the argument either being an expression or a type. 
https://go.dev/ref/spec#Allocation

PR #679, opened 2026-09-05
mutate struct field using the struct reference

kapad:
it may be unclear whether struct or struct pointers are mutable.

mutation via the struct reference and then printing via the pointer will add clarity.

### DIFF
--- a/examples/structs/structs.go
+++ b/examples/structs/structs.go
@@ -50,7 +50,7 @@ func main() {
 	fmt.Println(sp.age)
 
 	// Structs are mutable.
-	sp.age = 51
+	s.age = 51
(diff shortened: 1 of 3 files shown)


examples/structs/structs.go
```diff
@@ -50,7 +50,7 @@ func main() {
 	fmt.Println(sp.age)
 
 	// Structs are mutable.
-	sp.age = 51
```
eliben, line 53: I would maybe move the mutable part before the pointer part.

1. Show structs are mutable with `s`
2. Then in the "pointers" section, show both access and mutation one after another, through `sp`

kapad:
Feel free to close this PR and raise another instead.

